PR #343, opened 2019-01-04
Added realm version fixed #336

### LINKED ISSUE #336: Realm migrations

mistikk:


src/realm/realm.js
```diff
@@ -73,8 +73,24 @@ const authSchema = {
 const realm = new Realm({
   path: 'esteem.realm',
   schema: [userSchema, authSchema, draftSchema, settingsSchema, applicationSchema, scAccounts],
+  schemaVersion: 0,
```
feruzm: Adding schema and migration function won't cause issue for first alpha version?

mistikk: I tried with new scheme, and also I tried different scenario, it is working good

feruzm: nice, let's merge it then

PR #352, opened 2019-01-05
created leaderboard feature

ue:


feruzm:
screenshot would be really helpful

PR #384, opened 2019-01-09
Fixed login issue, changed pin code structure. Also fixed #331 fixed #332

### LINKED ISSUE #331: When you add account, it replaces the previous logged in account

### LINKED ISSUE #332: Users can't sign in with posting and active key

mistikk:
fixed #331 
fixed #332


src/providers/steem/auth.js
```diff
@@ -33,23 +32,23 @@ export const login = async (username, password) => {
...
     );
   }
-
   // Public keys of user
   const publicKeys = {
-    active: account.active.key_auths.map(x => x[0]),
-    memo: account.memo_key,
-    owner: account.owner.key_auths.map(x => x[0]),
-    posting: account.posting.key_auths.map(x => x[0]),
+    activeKey: account.active.key_auths.map(x => x[0])[0],
```
feruzm: hmm have you tested this with different accounts? Because some accounts have multiple key_auths and index [0][0] might catch different publickey. for example: https://steemd.com/@hellosteem

src/screens/application/container/applicationContainer.js
```diff
@@ -255,6 +291,7 @@ const mapStateToProps = state => ({
   currentAccountUsername: state.account.currentAccount.name,
   otherAccounts: state.account.otherAccounts,
   pinCode: state.account.pin,
+  currentAccountAuthType: state.account.currentAccount.local,
```
ue: lets get just crrentAccount from state because its calling twice time for name and also local thing.

src/screens/application/container/applicationContainer.js
```diff
@@ -102,28 +106,52 @@ class ApplicationContainer extends Component {
...
         });
       }
     });
 
-    if (realmData) {
-      await getUser(currentUsername)
+    if (realmData.length > 0) {
+      const realmObject = realmData.filter(data => data.username === currentUsername);
+
+      if (realmObject.length <= 0) {
```
ue: just change it `realmObject.lenght === 0`

src/screens/application/container/applicationContainer.js
```diff
@@ -102,28 +106,52 @@ class ApplicationContainer extends Component {
...
+                && !accountData.memoKey
+              ) {
+                realmData.splice(index, 1);
+                if (realmData.length <= 0) {
+                  dispatch(login(false));
+                  dispatch(logoutDone());
+                  removePinCode();
+                  setAuthStatus({ isLoggedIn: false });
+                  setExistUser(false);
+                  if (accountData.authType === 'steemConnect') {
```
ue: use constant from `constants/authType`

src/screens/application/container/applicationContainer.js
```diff
@@ -102,28 +106,52 @@ class ApplicationContainer extends Component {
...
+            userData.forEach((accountData, index) => {
+              if (
+                !accountData.accessToken
+                && !accountData.masterKey
+                && !accountData.postingKey
+                && !accountData.activeKey
+                && !accountData.memoKey
+              ) {
+                realmData.splice(index, 1);
+                if (realmData.length <= 0) {
```
ue: same as follow

src/providers/steem/auth.js
```diff
@@ -128,133 +138,70 @@ export const loginWithSC2 = async (code) => {
...
-      return response;
-    });
+  try {
+    await setPinCode(data.pinCode);
+    const users = await getUserData();
+    if (users.length > 0) {
+      users.forEach(async (userData) => {
+        if (userData.authType === 'masterKey') {
+          data.password = decryptKey(userData.masterKey, data.oldPinCode);
+        } else if (userData.authType === 'steemConnect') {
```
ue: can use constant

src/providers/steem/auth.js
```diff
@@ -128,133 +138,70 @@ export const loginWithSC2 = async (code) => {
...
-
-      await setAuthStatus(authData);
-      return response;
-    });
+  try {
+    await setPinCode(data.pinCode);
+    const users = await getUserData();
+    if (users.length > 0) {
+      users.forEach(async (userData) => {
+        if (userData.authType === 'masterKey') {
```
ue: can use constant

src/providers/steem/auth.js
```diff
@@ -128,133 +138,70 @@ export const loginWithSC2 = async (code) => {
...
-        currentUsername: userData.username,
-      };
-
-      await setAuthStatus(authData);
-      return response;
-    });
+  try {
+    await setPinCode(data.pinCode);
+    const users = await getUserData();
+    if (users.length > 0) {
```
ue: users can be return as nulll in that case this line goes crush

src/providers/steem/auth.js
```diff
@@ -116,9 +120,15 @@ export const loginWithSC2 = async (code) => {
     }
 
     setUserData(userData)
-      .then(() => {
+      .then(async () => {
         account.account.username = account.account.name;
```
ue: lets changed first name acccount.accont seems weird

src/providers/steem/dsteem.js
```diff
@@ -735,9 +731,8 @@ export const postContent = async (
 // Re-blog
 export const reblog = async (account, pinCode, author, permlink) => {
   const pin = getDigitPinCode(pinCode);
-
-  if (account.local.authType === AUTH_TYPE.MASTER_KEY) {
-    const key = decryptKey(account.local.postingKey, pin);
+  const key = getAnyPrivateKey(account.local, pin);
+  if (account.local.authType !== AUTH_TYPE.STEEM_CONNECT && key) {
```
ue: what if key doesn't exist this should be return something
```
if(!key){
return null;
}
```
or something like that. instead both check also it can return

ue: @mistikk

PR #446, opened 2019-01-17
Feature/share

ue:
### Issue number
#356


src/constants/options/post.js
```diff
@@ -1,3 +1,3 @@
-export default ['COPY LINK', 'REBLOG', 'REPLY'];
+export default ['COPY LINK', 'REBLOG', 'REPLY', 'SHARE'];
 
-export const VALUE = ['copy', 'reblog', 'reply'];
+export const VALUE = ['copy', 'reblog', 'reply', 'share'];
```
mistikk: I think we should move to language files

ue: updated for language

Issue #373, opened 2019-01-07
Suggestion: Collapse is needed for long comment


horpeyloaded:
#### Collapse is needed for comments like the below pix
![](https://i.ibb.co/42SYqPR/Screenshot-2019-01-05-22-46-36.png)

PR #452, opened 2019-01-17
Bugfix/#373

ue:
### What does this PR?
added a button its showing child all comments

### Steps to reproduce
just open a post and scroll to comments section
### Issue number
#373


feruzm:
nice 👍

feruzm:
If properly implemented, I think this should increase performance of post opening page, because you would only fetch top comments and all sub-comments are fetched only when user clicks on toggle...

PR #509, opened 2019-01-25
fixed upvote g bugs

ue:
### What does this PR?
changed upvote algorithm added global props to redux store


mistikk:
conflict @ue

ue:
fixed @mistikk

PR #533, opened 2019-02-05
added german hungarian indonesia russian etc.

ue:


feruzm:
@ue please make sure to enable languages that has 100% proofread... 
I just proofread Indonesian... checking other languages and we can add Acehnese as well, I will check that now...

image

feruzm:
ok, i have proofread all languages you have enabled, and new Acehnese as well, please enable that one as well... and now syncing results with github

feruzm:
fixed conflicts, only add `acehnese language files` and we are good to go and merge this...

PR #539, opened 2019-02-05
Fixed #468

### LINKED ISSUE #468: [eSteem Mobile] 2.0.6 [BUG] Incorrect markdown cursor position on Bold and Italic text in editor
Repository: https://github.com/esteemapp/esteem-mobile

**Describe the bug**
The editor in the mobile version of the eSteem app offers the possibility to insert text via click. When inserting a bold text, the cursor jumps to the end and creates the command ``****|`` which is very difficult to use. 

It would be better if the cursor were in the middle of the two asterisks to continue writing. There are two possibilities you can or should change here

- 1  After clicking on the Bold icon the following is inserted with the cursor position:
``**|**`` to also write a Bold Text

- 2 the first click on the Bold icon will insert the first two asterisks
``**|`` and after writing a text and clicking on the icon again 2 more asterisks ``**`` are added to do justice to the markdown.

With variant 2 it already works very well with the code snippet in the editor.

The same as above also happens with italicized text. When inserting an italic text ``**`` is inserted which is also a problem as described above. 

**To Reproduce**
1. Download the Latest Version of eSteem Mobile 2 App for Android
2. Go in to the Editor where you can write a Contribution
3. Hit the Bold or Italic Text Button to Insert the Markdown for these Text-Styles
4. There you see the Problem 

**Expected behavior**
My expectation was that if I wanted to insert a bold or italic text, the markdown would be placed correctly to ensure easy writing in the editor. When inserting the markdown command for code this works very well.

**Screenshots or Video**
Screen Video of Input in the eSteem Mobile Application
ezgif-4-ff36b1a6e0783123

At the eSteem Surfer Application is it working very well
ezgif-4-5a559b4674eb

 - Device: Huawei Mate 10 Pro
 - OS: Android 9
 - Version 2.0.6 latest

**Additional context**
In my personal opinion, it would be appropriate to change it like the Code Snippet, which has to be pressed twice on the Bold or Italic button. This is only my own opinion how to solve it best.

mistikk:
### Issue number
Fixed #468

### DIFF
--- a/src/components/markdownEditor/view/formats/applyListFormat.js
+++ b/src/components/markdownEditor/view/formats/applyListFormat.js
@@ -26,7 +26,7 @@ export default ({ getState, item, setState }) => {
 
   setState({ text: newText }, () => {
     setTimeout(() => {
-      setState({ selection: newSelection });
+      setState({ newSelection });
     }, 300);
   });
 };
--- a/src/components/markdownEditor/view/formats/applyWebLinkFormat.js
+++ b/src/components/markdownEditor/view/formats/applyWebLinkFormat.js
@@ -36,6 +36,6 @@ export default ({
     };
   }
   setState({ text: newText }, () => {
-    setState({ selection: newSelection });
+    setState({ newSelection });
   });
 };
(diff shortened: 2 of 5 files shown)


mistikk:
@ue updated

PR #548, opened 2019-02-07
Performance enhancement

mistikk:


feruzm:
where is `config.js` file? I see package.json has config js pointer, is it auto generated or custom?

mistikk:
in main folder @feruzm

PR #600, opened 2019-02-15
Bugfix/deeplink

mistikk:
### Steps to reproduce
      // esteem://esteem/@good-karma/esteem-mobile-v2-android-release
      // https://steemit.com/esteem/@good-karma/esteem-mobile-v2-android-release
      // https://busy.org/@good-karma/esteem-mobile-v2-android-release
      // https://steempeak.com/esteem/@good-karma/esteem-mobile-v2-android-release

      // https://steempeak.com/@good-karma/
      // https://busy.org/@good-karma
      // esteem://@good-karma
      // https://steemit.com/@good-karma/

### DIFF
--- a/src/screens/root/container/rootContainer.js
+++ b/src/screens/root/container/rootContainer.js
@@ -50,7 +50,7 @@ const RootContainer = () => (WrappedComponent) => {
 
     _handleOpenURL = (event) => {
       this._handleDeepLink(event.url);
-    }
+    };
 
     _handleDeepLink = async (url) => {
       if (!url) return;
@@ -116,7 +116,8 @@ const RootContainer = () => (WrappedComponent) => {
       }
 
       if (profile || content) {
-        setTimeout(() => {
+        this.navigationTimeout = setTimeout(() => {
+          clearTimeout(this.navigationTimeout);
           navigation.navigate({
             routeName,
             params,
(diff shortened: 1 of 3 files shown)


src/screens/root/container/rootContainer.js
```diff
@@ -113,11 +116,13 @@ const RootContainer = () => (WrappedComponent) => {
       }
 
       if (profile || content) {
-        navigation.navigate({
-          routeName,
-          params,
-          key: permlink || author,
-        });
+        setTimeout(() => {
```
ue: why you added time out here?

feruzm: i suspect, without it redirection won't work properly because app will be still loading up and loose redirect url data if there is no timeout... but please test it without `timeout` as well to confirm if it is necessary or not.

ue:
please remove `setTimeout` when this unmounted

src/screens/root/container/rootContainer.js
mistikk: yes we should wait until global props loading, we should change all of global props loading structure, I solved with timeout, after we changed global props structure we will remove timeout

feruzm: Can you create GitHub issue for that because we need to make sure that structure is improved, also give context on what should be improved?

mistikk: okay

mistikk:
@ue updated

PR #606, opened 2019-02-18
created comment filter

ue:
### What does this PR?
order comments according to selected filter
### Steps to reproduce
go a post then scroll to bottom. you will see a dropdown filter change from dropdown if order changed u will see re ordered comments


mistikk:
Reputation and votes section not working as correctly @ue

ue:
@mistikk Updated

Issue #601, opened 2019-02-18
[eSteem mobile v2 ios] Share is working as reply and reply is working as reblog.


face2face-utopian:
**Describe the bug**
I was in my profile section. I clicked on three vertical dotted and it has the following features 
1. Copy link
2. Reply
3. Share
4. Add to bookmark.
 The number 1 feature is working ok, the Second one gives the option to reblog while it is for *Reply*.
The third one is Share but it works as *Reply* and the 4th one is working as *Share*.

**To Reproduce**
1. Go to 'Profile page'
2. Click on 'Three vertical dotted just right sides of your post'
3. Try to use the available functions

**Expected behavior**

**Screenshots or Video**
![img_e6415](https://user-images.githubusercontent.com/36129744/52951834-dbffaf80-33a8-11e9-94a2-8322725fd923.JPG)

**Smartphone (please complete the following information):**
 - Device: [ iPhone6]
 - OS: [e.g. iOS 12
 

**Additional context**

ue:
Could you please test and close if it okay.

Android Apk Download link: https://install.appcenter.ms/orgs/esteem.app/apps/esteem/distribution_groups/beta_testers
IOS Ipa Download link:
https://install.appcenter.ms/orgs/esteem.app/apps/esteem-1/distribution_groups/beta_testers

Thank you. @face2face-utopian

Issue #560, opened 2019-02-09
[BUG] Wrong Links on own Profile - Drop Down Links for Reply, Bookmark and Share


louis88-steemit:
**Describe the bug**
If you are logged in to the app and go to your own profile and try to perform various actions from a written post there, this will fail. If you click the drop-down button of a post you wrote yourself, you have the following choices:
- Copy the link to the post
- Reply to the post
- Share the post in another app
- Bookmark the article

If you press link to copy everything is fine. But if you reply to a post or press the other buttons, the correct will not happen. The following happens

- Reply is linked to the Reblog Feature (Which should also not available for own Posts)
- Share is linked to Reply to the Post
and
- Add to Bookmark is linked to the Share Feature.

**To Reproduce**
1. Log in to your Steem Account by using the latest Version of the eSteem Mobile App
2. Go to your own Profile
3. On the Upper Right Side of the Post Hit the three Points to Open the Menu
4. Try to Bookmark , Share or Reply your own Post

**Expected behavior**
If you call the options for your own post, I assume and expect that exactly the function is deposited, which is also displayed.

**Screenshots or Video**
Youtube Hosted Video / Screen Video that shows the Issue
https://www.youtube.com/watch?v=VfLTMgaDcDY

**Smartphone:**
 - Device: Huawei Mate 10 Pro
 - OS: Latest Andrpid Version
 - Version of eSteem: v.2.0.8, 839

**Additional context**
I think that this will also show the options that are available for other posts. (contributions of other users) and here no special reprogramming was carried out like with own contributions one deals.

feruzm:
Good find, we will check it out!

mistikk:
@feruzm fixed and distributed for initial testers

ue:
Could you please test and close if it okay.

Android Apk Download link: https://install.appcenter.ms/orgs/esteem.app/apps/esteem/distribution_groups/beta_testers
IOS Ipa Download link:
https://install.appcenter.ms/orgs/esteem.app/apps/esteem-1/distribution_groups/beta_testers

Thank you. @louis88-steemit

PR #607, opened 2019-02-19
Fixed post dropdown issue

mistikk:
### Issue number
#601 
#560

### DIFF
--- a/src/components/postDropdown/container/postDropdownContainer.js
+++ b/src/components/postDropdown/container/postDropdownContainer.js
@@ -52,11 +52,11 @@ class PostDropdownContainer extends PureComponent {
   };
 
   // Component Functions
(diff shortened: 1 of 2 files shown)


src/components/postDropdown/container/postDropdownContainer.js
```diff
@@ -52,11 +52,11 @@ class PostDropdownContainer extends PureComponent {
...
 
   // Component Functions
-  _handleOnDropdownSelect = async (index) => {
+  _handleOnDropdownSelect = async (index, item) => {
     const { content, dispatch, intl } = this.props;
 
-    switch (index) {
-      case '0':
+    switch (item) {
+      case 'COPY LINK':
```
ue: these are good catch 👍

PR #619, opened 2019-02-21
Fixed image upload error

mistikk:


ue:
@mistikk did you test on IOS ? as much as I know its related Andorid side.

Issue #527, opened 2019-02-02
[Feature suggestion esteem app v2] Please Add "Open with" feature to open other steemit link


face2face-utopian:
Hi eStem Team,
I would like to see a new feature in eSteem app and that is "open with". Actually, sometimes some bloggers mention a link of the previous post or another steemit link and when we click on that link using esteem app v2 it goes redirect and the link open in the browser so if we would have feature in latest app then we can open that link our app only using "open with* feature. I have tried to show you all an example here:

![img_6222](https://user-images.githubusercontent.com/36129744/52164292-144e8f00-2715-11e9-890e-b3f0f3e4a02c.PNG)

Thanks
@face2face

face2face-utopian:
Yes i said about only steemit link. Thanks for acceptance.

feruzm:
moving this to next release cycle, current one (2.0.8) is overdue

horpeyloaded:
internal parsing of links are not working.

ue:
Could you please test and close if it okay.

Android Apk Download link: https://install.appcenter.ms/orgs/esteem.app/apps/esteem/distribution_groups/beta_testers
IOS Ipa Download link:
https://install.appcenter.ms/orgs/esteem.app/apps/esteem-1/distribution_groups/beta_testers

Thank you. @face2face-utopian

PR #620, opened 2019-02-21
Fix for #527

mistikk:


src/components/postElements/body/view/postBodyView.js
```diff
@@ -47,9 +47,38 @@ class PostBody extends PureComponent {
...
+  };
+
+  _handleBrowserLink = async (url) => {
+    if (!url) return;
+
+    let author;
+    let permlink;
+    const { intl } = this.props;
+
+    if (
```
ue: we should handle these url at `markdownToHtml.js` we can change as our app url but for now I'll approve maybe next time we can move these to that file. just quick nitpick otherwise approved 👍

Issue #434, opened 2019-01-15
Esteem Mobile V2: Tag Not Working When Clicked


yemmighto:
###  Project Information

- Repository
https://github.com/esteemapp/esteem-mobile

### Project Name

Esteem Mobile V2

Publisher: Esteem Mobile app

#### What I Actually Expected 

When I saw a post, and click on the tag used I was expecting to be shown the other post that use this same tag

#### The error shown

When I click the tag option, it didn't show anything at all. And I had to click other tags which bring same result 

 ![image](https://img.esteem.ws/fhkzx4iqdw.jpg)

To Reproduce bug

- open your esteem V2 mobile app
- go to feed and click a post
- scroll to the tag option and click different tag to see the result 

#### Bug recording

https://youtu.be/DZ1zxFBIfYk

yemmighto:
noted thanks

PR #635, opened 2019-02-26
Bugfix/#434

mistikk:
### Issue number
#434


ue:
@mistikk has conflits

mistikk:
@ue updated

PR #650, opened 2019-02-28
ePoint System Implemented

ue:

### DIFF
--- a/src/config/ePoint.js
+++ b/src/config/ePoint.js
@@ -1,0 +1,11 @@
+import axios from 'axios';
+import Config from 'react-native-config';
+
+const search = axios.create({
+  baseURL: Config.BACKEND_URL,
+  headers: {
+    'User-Agent': Config.USER_AGENT,
+  },
+});
+
+export default search;
--- a/src/providers/esteem/ePoint.js
+++ b/src/providers/esteem/ePoint.js
@@ -1,0 +1,65 @@
+import ePointApi from '../../config/ePoint';
+
+export const userActivity = (us, ty, bl = '', tx = '') => {
+  const params = { us, ty };
+
+  if (bl) {
+    params.bl = bl;
+  }
+
+  if (tx) {
+    params.tx = tx;
+  }
+
+  return ePointApi.post('/usr-activity', params).then(res => res.data);
+};
+
+export const transfer = (sender, receiver, amount) => new Promise((resolve, reject) => {
+  ePointApi
+    .post('/transfer', {
+      se: sender,
+      re: receiver,
+      am: amount,
+    })
+    .then((res) => {
+      resolve(res.data);
+    })
+    .catch((error) => {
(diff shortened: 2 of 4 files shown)


ue:
.env file new data send from discord channel

mistikk:
can you add new data to env.example @ue

src/config/ePoint.js
```diff
@@ -0,0 +1,11 @@
...
+import Config from 'react-native-config';
+
+const search = axios.create({
+  baseURL: Config.BACKEND_URL,
+  headers: {
+    'User-Agent': Config.USER_AGENT,
+  },
+});
+
+export default search;
```
mistikk: we dont need to add new api file for that, because we are using same backend url in here. My suggestion is

const headers: {
    'User-Agent': Config.USER_AGENT,
  }
ePointApi.post('/usr-activity', params, {headers: headers})

we can use this code inside of ePoint provider

src/providers/esteem/ePoint.js
```diff
@@ -0,0 +1,65 @@
+import ePointApi from '../../config/ePoint';
+
+export const userActivity = (us, ty, bl = '', tx = '') => {
```
mistikk: can you change variable names with long form

src/providers/esteem/ePoint.js
```diff
@@ -0,0 +1,65 @@
...
+
+  if (bl) {
+    params.bl = bl;
+  }
+
+  if (tx) {
+    params.tx = tx;
+  }
+
+  return ePointApi.post('/usr-activity', params).then(res => res.data);
```
mistikk: we should use catch, if api get error, app will crash right now

src/providers/steem/dsteem.js
```diff
@@ -376,14 +376,18 @@ export const getPostWithComments = async (user, permlink) => {
...
-// export const getAccountRC = username => client.call('rc_api', 'find_rc_accounts', { accounts: [username] });
-
 /**
  * @method upvote upvote a content
  * @param vote vote object(author, permlink, voter, weight)
  * @param postingKey private posting key
  */
-export const vote = async (currentAccount, pin, author, permlink, weight) => {
+
+export const vote = (account, pin, author, permlink, weight) => _vote(account, pin, author, permlink, weight).then((resp) => {
```
mistikk: can you write with seperate lines, it is getting max line length error

src/providers/steem/dsteem.js
```diff
@@ -742,7 +776,13 @@ export const postContent = async (
 };
 
 // Re-blog
-export const reblog = async (account, pinCode, author, permlink) => {
+
+export const reblog = (account, pinCode, author, permlink) => _reblog(account, pin, author, permlink).then((resp) => {
```
mistikk: eslint max length error

PR #739, opened 2019-04-01
created owner indicator

ue:

### DIFF
--- a/src/screens/application/container/applicationContainer.js
+++ b/src/screens/application/container/applicationContainer.js
@@ -1,7 +1,5 @@
 import React, { Component } from 'react';
-import {
-  Platform, BackHandler, Alert, NetInfo,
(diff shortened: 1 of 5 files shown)


src/screens/application/container/applicationContainer.js
```diff
@@ -170,19 +166,19 @@ class ApplicationContainer extends Component {
     let realmData = [];
     let currentUsername;
 
-    await getAuthStatus().then((res) => {
+    await getAuthStatus().then(res => {
       ({ currentUsername } = res);
       if (res) {
-        getUserData().then(async (userData) => {
+        getUserData().then(async userData => {
```
mistikk: this async code not necessary in here, we are not using await inside of func

mistikk:
@ue conflict

PR #870, opened 2019-05-23
Bugsnag integration 

mistikk:


ue:
@mistikk has a lot of conflits

ue:
@mistikk less conflits

PR #905, opened 2019-06-09
Fixed comment issues

mistikk:

### DIFF
--- a/src/components/comment/view/commentView.js
+++ b/src/components/comment/view/commentView.js
@@ -1,5 +1,5 @@
 import React, { PureComponent, Fragment } from 'react';
-import { View, Text } from 'react-native';
+import { View, Text, TouchableWithoutFeedback } from 'react-native';
 import ActionSheet from 'react-native-actionsheet';
 import { injectIntl } from 'react-intl';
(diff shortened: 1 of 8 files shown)


src/components/comment/view/commentView.js
```diff
@@ -56,115 +57,121 @@ class CommentView extends PureComponent {
...
-                  iconType="MaterialIcons"
-                  isClickable
-                  iconStyle={styles.iconStyle}
-                  iconSize={16}
-                  onPress={() => this._showSubCommentsToggle()}
-                  text={!isPressedShowButton ? `${comment.children} more replies` : ''}
-                />
-              </View>
+      <TouchableWithoutFeedback onLongPress={handleOnLongPress}>
+        <View>
```
ue: remove

ue: did you changed latest view tag but first one is still there is it correct or I am seeing wrong tag?

mistikk: we should user just 1 tag inside of TouchableWithoutFeedback, this is why I didnt remove view tag

PR #912, opened 2019-06-18
Bugfix/comment vote

ue:


src/components/comments/container/commentsContainer.js
```diff
@@ -122,14 +121,20 @@ class CommentsContainer extends Component {
...
+      currentAccount: { name },
+    } = this.props;
 
-    getComments(author, permlink).then(comments => {
-      this.setState({
-        comments,
-      });
-    });
+    await getComments(author, permlink, name)
+      .then(async comments => {
```
mistikk: can you remove async keyword, because you didnt use await inside of this function

src/providers/steem/dsteem.js
```diff
@@ -343,6 +353,8 @@ export const getPost = async (author, permlink, currentUserName = null) => {
   try {
     const post = await client.database.call('get_content', [author, permlink]);
 
+    const ugur = await getState(`/${post.category}/@${post.author}/${post.permlink}`);
+    console.log(ugur);
```
mistikk: ugur? :smile:

src/screens/post/container/postContainer.js
```diff
@@ -45,11 +49,10 @@ class PostContainer extends Component {
 
   componentWillReceiveProps(nextProps) {
     const { navigation } = this.props;
-    const { isFetch: nextIsFetch } =
-      nextProps.navigation.state && nextProps.navigation.state.params;
+    const { isFetch: nextIsFetch } = get(nextProps, 'navigation.state.params');
 
     if (nextIsFetch) {
-      const { author, permlink } = navigation.state && navigation.state.params;
+      const { author, permlink } = get(navigation, 'state.params');
```
mistikk: I think in here we should user like get(navigation.state, 'params'); because in prev line we are checking state value not navigation

ue: same thing

src/providers/steem/dsteem.js
ue: Sorry I forgot to remove god mode variable 🤣

ue:
@mistikk updated

PR #915, opened 2019-06-20
Removed reblog option from reblogged posts

mistikk:


src/screens/profile/screen/profileScreen.js
```diff
@@ -206,6 +206,7 @@ class ProfileScreen extends PureComponent {
                 tag={username}
                 key={username}
                 handleOnScroll={this._handleOnScroll}
+                hideReblogOption
```
ue: if the variable include boolean. lets change the name as `isHideReblogOption` or `isReblogOptionHide` its seems more valuable isn't it?

PR #933, opened 2019-07-07
created launch screen

ue:


ue:
@mistikk I am thinking add lottie animation in feature. that's why I removed styles for temporarily

PR #962, opened 2019-07-22
Bugfix/feed caching

mistikk:

### DIFF
--- a/src/components/posts/container/postsContainer.js
+++ b/src/components/posts/container/postsContainer.js
@@ -1,5 +1,6 @@
 import React, { PureComponent, Fragment } from 'react';
 import { connect } from 'react-redux';
+import get from 'lodash/get';
 
 // Component
 import PostsView from '../view/postsView';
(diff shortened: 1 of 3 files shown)


src/components/posts/container/postsContainer.js
```diff
@@ -53,7 +53,7 @@ class PostsContainer extends PureComponent {
     return (
       <PostsView
         handleOnScrollStart={this._handleOnScrollStart}
-        currentAccountUsername={currentAccount && currentAccount.username}
+        currentAccountUsername={currentAccount && (currentAccount.username || currentAccount.name)}
```
ue: can u use get from lodash also name is enoguht

src/redux/reducers/accountReducer.js
```diff
@@ -45,7 +45,9 @@ export default function(state = initialState, action) {
     case ADD_OTHER_ACCOUNT:
       return {
         ...state,
-        otherAccounts: [...state.otherAccounts, action.payload],
+        otherAccounts: state.otherAccounts.some(item => item.username === action.payload.username)
```
ue: u can use get also here

mistikk:
@ue updated

PR #963, opened 2019-07-22
Fixed transfer bugs

mistikk:

### DIFF
--- a/src/containers/transferContainer.js
+++ b/src/containers/transferContainer.js
@@ -152,6 +152,7 @@ class TransferContainer extends Component {
       })
       .catch(err => {
         dispatch(toastNotification(err.message));
+        navigation.goBack();
       });
   };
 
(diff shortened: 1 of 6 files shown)


src/containers/transferContainer.js
```diff
@@ -152,6 +152,7 @@ class TransferContainer extends Component {
       })
       .catch(err => {
         dispatch(toastNotification(err.message));
+        navigation.goBack();
```
ue: can you please move to up line

src/redux/reducers/uiReducer.js
```diff
@@ -7,7 +7,7 @@ import {
 const initialState = {
   activeBottomTab: 'HomeTabbar',
   isCollapsePostButton: false,
-  toastNotifcaion: '',
+  toastNotification: '',
```
ue: good catch

mistikk:
@ue updated

PR #1005, opened 2019-08-01
Upgrade app center libraries

mistikk:


feruzm:
@mistikk have you checked that it won't break working parts? namely push notifications, etc.

mistikk:
I checked just push notification, other libraries require release build, we can check in last testing with testflight @feruzm

PR #1013, opened 2019-08-05
typo locale fix and todo: move toast messages to locales

feruzm:


feruzm:
@mistikk we need to check all toast messages as well...

PR #1033, opened 2019-08-10
Feature/leaderboard filter

ue:

### DIFF
--- a/src/constants/options/leaderboard.js
+++ b/src/constants/options/leaderboard.js
@@ -1,0 +1,3 @@
+export default ['day', 'week', 'month'];
+
+export const VALUE = ['Daily', 'Weekly', 'Montly'];
--- a/src/constants/options/leaderboard.js
+++ b/src/constants/options/leaderboard.js
@@ -1,3 +1,3 @@
 export default ['day', 'week', 'month'];
 
-export const VALUE = ['Daily', 'Weekly', 'Montly'];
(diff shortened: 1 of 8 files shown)


src/constants/options/leaderboard.js
```diff
@@ -0,0 +1,3 @@
+export default ['day', 'week', 'month'];
+
+export const VALUE = ['Daily', 'Weekly', 'Montly'];
```
mistikk: can you move to language file

src/providers/esteem/esteem.js
```diff
@@ -131,7 +131,10 @@ export const addFavorite = (currentUsername, targetUsername) =>
 export const removeFavorite = (currentUsername, targetUsername) =>
   api.delete(`/favoriteUser/${currentUsername}/${targetUsername}`);
 
-export const getLeaderboard = () => api.get('/leaderboard').then(resp => resp.data);
+export const getLeaderboard = duration =>
+  api.get('/leaderboard', { params: { duration } }).then(resp => {
+    return resp.data;
+  });
```
mistikk: can you add catch callback and you can send to bugsnag

src/constants/options/leaderboard.js
feruzm: Also make sure spelling is correct, “Monthly”, sublime has spell checker plugin if you use sublime 😉

ue:
@mistikk updated

PR #1056, opened 2019-08-21
Feature/downvote

mistikk:

### DIFF
--- a/src/components/upvote/container/upvoteContainer.js
+++ b/src/components/upvote/container/upvoteContainer.js
@@ -53,12 +53,12 @@ class UpvoteContainer extends PureComponent {
     } = this.props;
 
     const author = get(content, 'author');
-    const isVoted = get(content, 'is_voted')
-      ? parseInt(get(content, 'is_voted'), 10) / 10000
-      : false;
-    const isDownVoted = get(content, 'is_down_voted')
-      ? (parseInt(get(content, 'is_down_voted', false), 10) / 10000) * -1
-      : false;
+    const isVoted =
+      get(content, 'is_voted', false) && parseInt(get(content, 'is_voted'), 10) / 10000;
+    const isDownVoted =
+      get(content, 'is_down_voted', false) &&
+      (parseInt(get(content, 'is_down_voted'), 10) / 10000) * -1;
+
     const totalPayout = get(content, 'total_payout');
     const isDecinedPayout = get(content, 'is_declined_payout');
     const permlink = get(content, 'permlink');
--- a/src/components/upvote/container/upvoteContainer.js
+++ b/src/components/upvote/container/upvoteContainer.js
@@ -53,7 +53,12 @@ class UpvoteContainer extends PureComponent {
     } = this.props;
 
     const author = get(content, 'author');
-    const isVoted = get(content, 'is_voted');
+    const isVoted = get(content, 'is_voted')
+      ? parseInt(get(content, 'is_voted'), 10) / 10000
+      : false;
+    const isDownVoted = get(content, 'is_down_voted')
+      ? (parseInt(get(content, 'is_down_voted', false), 10) / 10000) * -1
+      : false;
     const totalPayout = get(content, 'total_payout');
     const isDecinedPayout = get(content, 'is_declined_payout');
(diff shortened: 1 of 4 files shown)


src/components/upvote/container/upvoteContainer.js
```diff
@@ -53,7 +53,12 @@ class UpvoteContainer extends PureComponent {
     } = this.props;
 
     const author = get(content, 'author');
-    const isVoted = get(content, 'is_voted');
+    const isVoted = get(content, 'is_voted')
```
ue: in short case;

`const isVoted = get(content, 'is_voted', false)
      && parseInt(get(content, 'is_voted'), 10) / 10000;`

src/components/upvote/container/upvoteContainer.js
```diff
@@ -53,7 +53,12 @@ class UpvoteContainer extends PureComponent {
...
+      ? parseInt(get(content, 'is_voted'), 10) / 10000
+      : false;
+    const isDownVoted = get(content, 'is_down_voted')
```
ue: both

src/components/upvote/view/upvoteView.js
```diff
@@ -156,9 +221,10 @@ class UpvoteView extends Component {
       iconType = 'AntDesign';
     }
 
-    const _percent = `${(sliderValue * 100).toFixed(0)}%`;
+    const _percent = `${downvote ? '-' : ''}${(sliderValue * 100).toFixed(0)}%`;
     const _amount = `$${amount}`;
     const _totalPayout = totalPayout || '0.000';
+    const _sliderColor = downvote ? '#ec8b88' : '#357ce6';
```
ue: there is no `sliderColor` underscore not necessary

src/components/upvote/view/upvoteView.js
```diff
@@ -183,6 +249,14 @@ class UpvoteView extends Component {
                       isShow={!isVoting}
                     />
                   </View>
+                ) : isDownVoted ? (
+                  <Icon
```
ue: can merge with follow icon like inside the props can use is down voted

ue: write to me from dm if the comment not explainable

src/utils/postParser.js
```diff
@@ -144,5 +148,22 @@ export const parseComments = async (comments, currentUserName) => {
...
+  );
+  if (result) {
+    return result.percent;
+  }
+  return false;
+};
+
+const isDownVoted = (activeVotes, currentUserName) => {
+  const result = activeVotes.find(
+    element => element.voter === currentUserName && element.percent < 0,
```
ue: can use `get`

ue:
@mistikk when pressing the downvote button then changing slide the price value changing I don't thing so if its necessary am I right @feruzm ?

src/components/upvote/container/upvoteContainer.js
mistikk: this can be wrong line?

src/components/upvote/view/upvoteView.js
feruzm: On video Mustafa post, I saw the color, maybe on mobile it is not visible?!

feruzm:
@ue we want user to push downvote button twice, that's why different value/slider on first click and casting downvote on second click is good way.

https://cdn.discordapp.com/attachments/487610005558657034/613810302227709973/Screen_Recording_2019-08-21_at_21.52.57.mov

mistikk:
@ue updated

src/utils/postParser.js
```diff
@@ -144,5 +148,22 @@ export const parseComments = async (comments, currentUserName) => {
   return _comments;
 };
 
-const isVoted = (activeVotes, currentUserName) =>
-  activeVotes.some(v => v.voter === currentUserName && v.percent > 0);
+const isVoted = (activeVotes, currentUserName) => {
+  const result = activeVotes.find(
+    element => element.voter === currentUserName && element.percent > 0,
```
ue:
```suggestion
    element => get(element, 'voter') === currentUserName && get(element, 'percent', 0) > 0,
```

PR #1058, opened 2019-08-22
Bugfix/promote post

ue:


feruzm:
So now, How is it handling if promoted post and feed has same post in list?

PR #1119, opened 2019-08-29
Fixed transaction expired issue

mistikk:


src/providers/steem/dsteem.js
```diff
@@ -489,6 +489,10 @@ const _vote = async (currentAccount, pin, author, permlink, weight) => {
           resolve(result);
         })
         .catch(err => {
+          if (err.jse_info && err.jse_info.code === 4030100) {
```
ue: use get @mistikk

PR #1147, opened 2019-09-09
Fixed sc login issue

mistikk:

### DIFF
--- a/src/providers/steem/auth.js
+++ b/src/providers/steem/auth.js
@@ -128,7 +128,7 @@ export const loginWithSC2 = async (code, isPinCodeOpen) => {
       postingKey: '',
       activeKey: '',
       memoKey: '',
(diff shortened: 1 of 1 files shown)


feruzm:
what issue does it solve? I thought all sc issues are gone, because nobody reported any more bugs unless this is fix for old bug report

mistikk:
it is about authority issue. If login with steemconnect as 2, 3.. account (without first account) it throws authoriy error

feruzm:
oh you mean if first account logout/removed then all other accounts 2,3 won't work?!

feruzm:
If this is the case, yes we need to merge it after testing, I am sure Ugur will do some tests and I will check on device after testflight version.

mistikk:
@feruzm when login with sc as first account, you can not see this issue. when you add a new account with sc login, you will see this error

PR #1148, opened 2019-09-10
Sc upgrade

mistikk:


feruzm:
quite big PR... looks good to me, but would suggest to branch off from latest master always

mistikk:
it is about, difference between master and development, i fixed it @feruzm

PR #1168, opened 2019-09-24
Enhancment/redeem

ue:


src/config/locales/en-US.json
```diff
@@ -222,7 +222,8 @@
     "fetch_error": "Fetching data failed, please try again or notify us at info@esteem.app",
     "connection_fail": "Connection Failed!",
     "connection_success": "Successfully connected!",
-    "checking": "Cheking..."
+    "checking": "Cheking...",
```
feruzm: should be Checking..., spelling error

PR #1172, opened 2019-09-25
created new boost container

ue:


src/screens/boost/screen/boostScreen.js
```diff
@@ -20,9 +23,7 @@ class BoostScreen extends PureComponent {
 
   constructor(props) {
     super(props);
-    this.state = {
-      selectedBoost: 0,
-    };
+    this.state = {};
```
mistikk: change with functional component

PR #1173, opened 2019-09-26
Fixed percent bar and cover image issues

mistikk:


src/containers/profileContainer.js
```diff
@@ -339,9 +339,9 @@ class ProfileContainer extends Component {
     let votingPower;
     let resourceCredits;
 
-    if (selectedUser) {
-      votingPower = getVotingPower(selectedUser).toFixed(1);
-      resourceCredits = getRcPower(selectedUser).toFixed(1);
+    if (user) {
+      votingPower = getVotingPower(user).toFixed(1);
+      resourceCredits = getRcPower(user).toFixed(1);
```
ue: can u remove `selectedUser` if it not using anywhere

mistikk: it is using inside of screen component

PR #1175, opened 2019-09-28
v2.2.3

mistikk:


feruzm:
This should be v2.2.3 if merged, so make sure before merge version updates... or we can later include more commits from development branch before merging

mistikk:
This was for last version. I want to fix last bugs in latest version and push with codepush. master should has latest version of production

feruzm:
If it only fixes bugs without new features, we can merge it as 2.2.3 because I don't want to use codepush that causes some unstability for startup...

PR #1181, opened 2019-10-01
Added information alert for old google play version error

mistikk:


ue:
@mistikk fix confilits

PR #1209, opened 2019-10-12
Typo: vote_perecent to vote_percent

ausbitbank:
### What does this PR do ?
Just a simple typo fix for a variable that I randomly spotted in postParser.js. I have very little knowledge of the overall esteem codebase, so this should be reviewed but I'm 99% sure this is just a typo.


ue:
Thank you for your support @ausbitbank

PR #1220, opened 2019-10-16
Changed error message for wrong time

mistikk:


src/providers/steem/dsteem.js
```diff
@@ -1213,8 +1219,11 @@ export const profileUpdate = async (params, pin, currentAccount) => {
         })
         .catch(error => {
           if (get(error, 'jse_info.code') === 4030100) {
-            error.message =
-              'We noticed that your device has incorrect date or time. Please fix Date & Time or Set Automatically and try again.';
+            error.message = `Your device time; ${
+              error.jse_info.stack[0].data['trx.expiration']
+            } \n Current time; ${
+              error.jse_info.stack[0].data.now
+            } \n \n We noticed that your device has incorrect date or time. Please fix Date & Time or Set Automatically and try again.`;
```
ue: can u please write a util for that

mistikk: i think util not necessary for that, because it will just using on this class, I can create a new function on dsteem.js, what do you think about that @ue

ue: I prefer utils cuz dsteem not quitly related with that its doing dsteem thing.

mistikk:
@ue updated

PR #1222, opened 2019-10-16
Enhancment/darktheme

ue:


mistikk:
@ue when we disable to darkmode on settings, it is going again enable after kill to app

ue:
@mistikk  Fixed

ue:
ready

PR #1223, opened 2019-10-17
[Feature] Open post and profile links on search bar

mistikk:

### DIFF
--- a/src/utils/postUrlParser.js
+++ b/src/utils/postUrlParser.js
@@ -1,0 +1,60 @@
+export default url => {
+  const parseCatAuthorPermlink = u => {
+    const r = /^https?:\/\/(.*)\/(.*)\/(@[\w.\d-]+)\/(.*)/i;
+    const match = u.match(r);
+    if (match && match.length === 5) {
+      return {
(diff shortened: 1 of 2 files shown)


src/utils/postUrlParser.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    return null;
+  };
+
+  if (
+    ['https://esteem.app', 'https://steemit.com', 'https://steempeak.com'].some(x =>
+      url.startsWith(x),
+    )
+  ) {
```
feruzm: What’s difference here, parseCatAuthorpermlink and parseAuthorpermlink?

mistikk: parseCatAuthorpermlink for baseUrl + category + author + permlink
parseAuthorpermlink for baseUrl + author + permlink.

We are using same code with surfer side

feruzm: oh I see 👍

ue:
@mistikk it should be work as search bar.
shuld not open the gift right after that. pasting the url. should be get as suggestion bottom of the input then user should press if the url matching some of post or post

PR #1235, opened 2019-10-28
Refactor/markdown

mistikk:


ue:
its getting freze in ios did u test in ios? @mistikk

mistikk:
@ue yes tested on both of platform. What is the error in debugger

mistikk:
@ue updated

PR #1238, opened 2019-10-29
Fixed auto navigate action after login, changed navigation actions wi…

mistikk:
…th navigation service


src/screens/application/container/applicationContainer.js
```diff
@@ -38,6 +38,7 @@ import {
 import { getUser, getPost } from '../../../providers/steem/dsteem';
 import { switchAccount } from '../../../providers/steem/auth';
 import { setPushToken } from '../../../providers/esteem/esteem';
+import NavigationService from '../../../navigation/service';
```
ue: why did u use as all func instead using like this `{ navigate }`

mistikk: for easy readability. when using navigate it can be mixed with navigation navigate method

ue: but you only used `navigate`

PR #1282, opened 2019-11-11
fixed spin count stlye issue

ue:


feruzm:
This is not merged into v2.2.3 I think

PR #1291, opened 2019-11-13
fixed login bug place holder etc.

ue:

### DIFF
--- a/src/components/posts/view/postsStyles.js
+++ b/src/components/posts/view/postsStyles.js
@@ -32,4 +32,7 @@ export default EStyleSheet.create({
     width: 193,
     height: 189,
   },
+  placeholderWrapper: {
+    flex: 1,
+  },
 });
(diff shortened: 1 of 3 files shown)


feruzm:
Not sure if this is relevant or not but upvote button style is changed since last couple updates from TestFlight. Why?! Maybe resizing wasn’t possible, but that shouldn’t be reason to change icon type...

ue:
@feruzm I fixed feed blog login to see issue and feed loading issues. And the other one is for the icon size  will be same for downvote and upvote.
with that pr
Screen Shot 2019-11-16 at 12 23 40
Screen Shot 2019-11-16 at 12 23 48


Screen Shot 2019-11-16 at 12 47 46
Screen Shot 2019-11-16 at 12 48 08
Screen Shot 2019-11-16 at 12 48 21

feruzm:
Ok, thanks Ugur! Lgtm

PR #1319, opened 2019-11-27
Steemconnect styling, some locale changes

feruzm:


src/screens/login/screen/loginScreen.js
```diff
@@ -190,7 +190,7 @@ class LoginScreen extends PureComponent {
           isFullScreen
           isCloseButton
           handleOnModalClose={this._handleOnModalToggle}
-          title="Steemconnect Login"
+          title="Sign in"
```
mistikk: we can move to language file

feruzm: yes 👍

PR #1322, opened 2019-11-27
Bugfix/webview

mistikk:


mistikk:
yes we forgot to enable it   😄

PR #1324, opened 2019-11-27
New/wallet

ue:


src/components/basicUIElements/view/placeHolder/boostPlaceHolderView.js
```diff
@@ -20,7 +20,7 @@ const BoostPlaceHolder = () => {
         {({ isDarkTheme }) => {
           const color = isDarkTheme ? '#2e3d51' : '#f5f5f5';
           return (
-            <View style={styles.container} key={i}>
+            <View style={styles.container} key={i.toString()}>
```
ue: this is a loader there is no particular information to apply key, thats why used index to do it

PR #1330, opened 2019-11-29
Bugfix/url scheme

mistikk:


feruzm:
There is some conflicts though

PR #1334, opened 2019-11-29
swiper nightly

feruzm:
Possible Fix for Android crash


feruzm:
@ue styling might be bit off maybe due to upgrade?!
Screenshot_1575052492

PR #1340, opened 2019-11-30
increase image sizes for better quality

feruzm:


feruzm:
Simple size changes, should be fine, but please see if it breaks some design patterns which I didn’t know about

PR #1353, opened 2019-12-04
WIP: Wallet improvements

feruzm:
improvements and functionality tests, don't merge yet

- [x] withdraw page should show saving balance

- [x] estimation on each token page should be on that specific balance estimation

- [x] pull down to refresh on all token pages

- [x]  delegate page, To field seem wrongly formatted

- [x] currency rate should choose user currency preference


src/utils/wallet.js
```diff
@@ -156,6 +157,14 @@ export const groomingWalletData = async (user, globalProps) => {
 
   walletData.estimatedValue = totalSteem * pricePerSteem + totalSbd;
 
+  const ppSbd = await getCurrencyTokenRate('usd', 'sbd');
```
feruzm: @mistikk Can you check this and possibly fix? I added currency rate for STEEM and SBD to get correct estimated value, but it is only getting `usd` rate and when user changes currency to EUR, it shows completely wrong estimated values.

feruzm:
@mistikk Please check this and possibly fix? I added currency rate for STEEM and SBD to get correct estimated value, but it is only getting usd rate and when user changes currency to EUR, it shows completely wrong estimated values.
1. Improve currency rate
2. Add pull down to refresh balances
3. Fix input size on Delegate page, `To` field input.

PR #1364, opened 2019-12-05
Added open post images feature

mistikk:


mistikk:
it has download feature when you long press to image, but i should test for that

feruzm:
we can test and do PR again if there is an issue 👍

horpeyloaded:
Don't forget `copy image direct link`, that will complete it 🙂

PR #1380, opened 2019-12-10
Wallet bug fixes

feruzm:
- [x] fix estimated sp value

- [x] show delegations

- [x] wallet header, balances won't refresh (it should refresh when pull down to refresh is performed.


feruzm:
@ue can you add balance refreshing effect? check details in above comment, only that issue is left to fix on this PR.

PR #1394, opened 2019-12-12
Fixed android white screen and crash issues

mistikk:


feruzm:
@mistikk Where is Pincode and Steemconnect screens, they are removed from `routes.js`, not need to define them?

mistikk:
we are using with modal to those screens, we just forgot to remove this screens from routes.js

feruzm:
ok, then let's merge this!

PR #1405, opened 2019-12-16
Fixed android save image issue

mistikk:


package.json
```diff
@@ -24,10 +24,10 @@
   "dependencies": {
     "@babel/runtime": "^7.5.5",
     "@esteemapp/esteem-render-helpers": "^1.2.2",
-    "@ptomasroos/react-native-multi-slider": "^1.0.0",
-    "@react-native-community/cameraroll": "^1.3.0",
     "@esteemapp/react-native-autocomplete-input": "^4.2.1",
     "@esteemapp/react-native-multi-slider": "^1.1.0",
+    "@ptomasroos/react-native-multi-slider": "^1.0.0",
```
feruzm: This is not necessary, please remove this package, we have `@esteemapp/react-native-multi-slider` that replaces this.

src/components/postElements/body/view/postBodyView.js
```diff
@@ -127,23 +128,73 @@ const PostBody = ({
...
+      Promise.resolve();
+    } catch (error) {
+      Promise.reject(error);
+    }
+  };
+
+  const _downloadImage = async uri => {
+    return RNFetchBlob.config({
+      fileCache: true,
+      appendExt: 'jpg',
```
feruzm: Do we have proper error handling if image meta type is different, like .gif or .png or .svg, etc.?

package.json
feruzm: on recent PR merge, I have removed it, so if you remove on this branch, it won't conflict

mistikk: okay i will remove it

src/components/postElements/body/view/postBodyView.js
mistikk: i tried with gif image it is working. Actually i added this config for an android bug

PR #1406, opened 2019-12-16
Fixed text input android context menu issue

mistikk:


package.json
```diff
@@ -24,10 +24,10 @@
   "dependencies": {
     "@babel/runtime": "^7.5.5",
     "@esteemapp/esteem-render-helpers": "^1.2.2",
-    "@ptomasroos/react-native-multi-slider": "^1.0.0",
-    "@react-native-community/cameraroll": "^1.3.0",
     "@esteemapp/react-native-autocomplete-input": "^4.2.1",
     "@esteemapp/react-native-multi-slider": "^1.1.0",
+    "@ptomasroos/react-native-multi-slider": "^1.0.0",
```
feruzm: also need removing

src/components/formInput/view/formInputView.js
```diff
@@ -13,123 +13,107 @@ import { getResizedAvatar } from '../../../utils/image';
...
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    const { isValid } = this.props;
-
-    if (nextProps.isValid !== isValid) {
-      this.setState({ isValid: nextProps.isValid });
-    }
-  }
+  // Workaround for android context (copy/paste) menu
+  useEffect(() => {
+    setValue(' ');
```
feruzm: why whitespace?

src/components/formInput/view/formInputView.js
```diff
@@ -13,123 +13,107 @@ import { getResizedAvatar } from '../../../utils/image';
...
+    setTimeout(() => {
```
feruzm: i guess this is required for context menu to resolve or something 🤔

package.json
```diff
@@ -80,8 +80,8 @@
     "react-native-webview": "^7.5.2",
     "react-navigation": "^4.0.10",
     "react-navigation-drawer": "^2.3.3",
-    "react-navigation-redux-helpers": "^2.0.8",
-    "react-navigation-stack": "1.10.3",
+    "react-navigation-redux-helpers": "^4.0.1",
```
feruzm: According to [this](https://github.com/react-navigation/redux-helpers/releases/tag/3.0.0) or later changelogs, don't we need `yarn add @react-navigation/core` ?

package.json
mistikk: sorry i will remove it

src/components/formInput/view/formInputView.js
mistikk: we should update to input value for resolve to context menu issue, this issue can due from react-navigation but i couldnt find any solution for that

src/components/formInput/view/formInputView.js
mistikk: it is a workaround solution

package.json
mistikk: i am not sure, i didnt see anything about that on readme. and it is working as expected right now

feruzm: ok

src/components/formInput/view/formInputView.js
feruzm: Btw, Do we have this problem in Editor ?

mistikk: it is strange because we solved to this problem on editor with `contextMenuHidden={false}` code, but it is not working on here

src/components/formInput/view/formInputView.js
feruzm: Yes, I also checked react-navigation before, tried package upgrades, didn't help anything. Something new or the way we are using it might be buggy. One more theory I have is that, it might be Modal component related.

`This is what I see on terminal`

  WARN  Warning: componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-async-component-lifecycle-hooks for details.
                                
 * Move data fetching code or side effects to componentDidUpdate.
 * If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://fb.me/react-derived-state
 * Rename componentWillReceiveProps to UNSAFE_componentWillReceiveProps to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.
                                
 Please update the following components: **ModalDropdown**, Popover

mistikk: this warn mostly related with 3th party libraries, we changed all componentWillReceiveProps function with UNSAFE_componentWillReceiveProps and we are refactoring with hooks. But some libraries didnt change yet

feruzm: Yes, read my comment on Discord. I have updated couple components locally so we don't have to wait some packages (some of them inactive). So we can fork, make change ourselves.

PR #1419, opened 2019-12-17
Fixed crash issue when user click to notification

mistikk:


src/screens/launch/screen/launchStyles.js
```diff
@@ -2,17 +2,21 @@ import EStyleSheet from 'react-native-extended-stylesheet';
 
 export default EStyleSheet.create({
   container: {
-    flex: 1,
-    justifyContent: 'center',
-    alignItems: 'center',
+    position: 'absolute',
```
feruzm: why style change is necessary?

mistikk: in the prev system; just launch screen shows until app is ready, but right now launch screen shows  on application screen. if you are asking for why i changed this, because error due from navigation initial config, we are setting initial config on application screen. in this case application screen not in the screen so initial config is not readt but we are trying to navigate a screen

mistikk: i am not sure i can explain it :)

feruzm: I got it... so basically notification data was being fed to Launch screen as oppose to Application screen. Styling to take entire application screen during that load. makes sense

PR #1421, opened 2019-12-17
fixed dark theme issues

ue:


src/screens/application/index.js
```diff
@@ -30,6 +30,10 @@ const Application = () => {
         isThemeReady,
         isPinCodeRequire,
       }) => {
+        if (showAnimation || !isReady || !isRenderRequire || !isThemeReady) {
```
feruzm: Mustafa moved this inside applicationscreen so that notification opening works, @mistikk, @ue

ue: otherwise darktheme goingto crush @feruzm

PR #1726, opened 2020-07-19
Updated firebase push notification library

mistikk:


ios/Podfile
```diff
@@ -35,11 +35,6 @@ target 'Ecency' do
   pod 'glog', :podspec => '../node_modules/react-native/third-party-podspecs/glog.podspec'
   pod 'Folly', :podspec => '../node_modules/react-native/third-party-podspecs/Folly.podspec'
   pod 'rn-fetch-blob', :path => '../node_modules/rn-fetch-blob'
-  # Add the Firebase pod for Google Analytics
```
feruzm, line 38: Why are we removing these from podfile? I see them in podfile.lock file, how is it possible if not in pod file they can be found in podfile.lock?

PR #1743, opened 2020-07-28
Feature/share

mistikk:


android/app/src/main/AndroidManifest.xml
```diff
@@ -64,6 +64,31 @@
...
+            <action android:name="android.intent.action.SEND" />
+            <category android:name="android.intent.category.DEFAULT" />
+            <data android:mimeType="image/*" />
+        </intent-filter>
+        <intent-filter>
+            <action android:name="android.intent.action.SEND_MULTIPLE" />
+            <category android:name="android.intent.category.DEFAULT" />
+            <data android:mimeType="image/*" />
+        </intent-filter>
+        <intent-filter>
```
feruzm, line 82: for now, we don't need video/* files because we don't have video uploads yet.

PR #1852, opened 2021-02-25
Posts Feed Optimisations - Experimental

noumantahir:
### What does this PR do?

based on my initial findings, the major stall app face is when ImageSize package us used by PostCardView and animated empty view holders.

1. I have removed the useEffect function from PostCardView and moved rest of simple calculation to main body. as ImageSize we'll have discuss either replacement or repositions of these calls.

2. Vote calculations are being done asynchronously now, also change the component type of UpvoteContainer to functional.

3. The animated empty view were taking too much toll on Android, I have commented them for now

### Suggestions/Questions?
1. We should return the image ratios with post payload, pulling them using thumbnail uri during runtime is too many calls. Is there any way to do that from BE?

2. list of active_votes is retuned as a long list in home feed that is essentially never needed, imo, a better approach would be to only return the total number of votes, and status vote related to user (from Backend). App will perform worst for most liked posts.

3. similarly complete body of post is returned in feed response, which again can be avoid and made a part of post details as an attempt to reduce payload size of feed call... what do you think.?

### DIFF
--- a/src/components/postCard/view/postCardView.js
+++ b/src/components/postCard/view/postCardView.js
@@ -38,11 +38,13 @@ const PostCardView = ({
   intl,
   activeVotes,
 }) => {
-  const [rebloggedBy, setRebloggedBy] = useState(get(content, 'reblogged_by[0]', null));
-  const [activeVot, setActiveVot] = useState(activeVotes);
-  const [calcImgHeight, setCalcImgHeight] = useState(300);
-  const [images, setImages] = useState({});
-  //console.log(activeVotes);
+  // const [rebloggedBy, setRebloggedBy] = useState(get(content, 'reblogged_by[0]', null));
+  // const [activeVot, setActiveVot] = useState(activeVotes);
+  // const [images, setImages] = useState({});
+  // const [calcImgHeight, setCalcImgHeight] = useState(300);
+  const calcImgHeight = 300;
+
+
   // Component Functions
 
   const _handleOnUserPress = () => {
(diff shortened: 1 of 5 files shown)


feruzm:
@noumantahir Sent invite, could you create branch inside repo and PR from there? It would help me to test changes quicker.

> We should return the image ratios with post payload, pulling them using thumbnail uri during runtime is too many calls. Is there any way to do that from BE?

As images are user submitted getting ratio of each thumbnail is only possible after image is downloaded, that's why we added very small thumbnail download to determine ratio/size. Not perfect solution I know, but that's what we have for now. One solution is to use same image after its downloaded and then show or get ratio and adjust.

>list of active_votes is retuned as a long list in home feed that is essentially never needed, imo, a better approach would be to only return the total number of votes, and status vote related to user (from Backend). App will perform worst for most liked posts.

There is `.stats.total_votes` for numbers as for status of vote related to `observer` is being [worked on](https://gitlab.syncad.com/hive/hivemind/-/issues/133)

>similarly complete body of post is returned in feed response, which again can be avoid and made a part of post details as an attempt to reduce payload size of feed call... what do you think.?

We used to have truncate_body parameter but because we show summary of post, we were not fully utilizing that. I will ask backend team to consider adding it back.

noumantahir:
> As images are user submitted getting ratio of each thumbnail is only possible after image is downloaded, that's why we added very small thumbnail download to determine ratio/size. Not perfect solution I know, but that's what we have for now. One solution is to use same image after its downloaded and then show or get ratio and adjust.

One solution I implemented in one of app, I would get image ratio before uploading images and save it's ratio along side the saved object... another better solution could be pure backend, a service that reads newly created post and process it's images to update post data.

We can let the image component to auto adjust height as it's loaded, but it will add a glitch for users. caching ratios after fetching feed should work for now yet with added load times.

PR #1853, opened 2021-02-25
Posts Feed Optimisations - Experimental

noumantahir:
### What does this PR do?

based on my initial findings, the major stall app face is when ImageSize package us used by PostCardView and animated empty view holders.

1. I have removed the useEffect function from PostCardView and moved rest of simple calculation to main body. as ImageSize we'll have discuss either replacement or repositions of these calls.

2. Vote calculations are being done asynchronously now, also change the component type of UpvoteContainer to functional.

3. The animated empty view were taking too much toll on Android, I have commented them for now


feruzm:
Trying to cast a vote gives this error.
Screenshot_1614327407

src/components/upvote/container/upvoteContainer.js
```diff
@@ -22,17 +22,35 @@ import UpvoteView from '../view/upvoteView';
...
+    const _isVoted = await isVotedFunc(activeVotes, get(currentAccount, 'name'));
+    const _isDownVoted = await isDownVotedFunc(activeVotes, get(currentAccount, 'name'));
+
+    setIsVoted(_isVoted && parseInt(_isVoted, 10) / 10000);
+    setIsDownVoted(_isDownVoted && (parseInt(_isDownVoted, 10) / 10000) * -1);
+  };
 
-  _setUpvotePercent = (value) => {
+  const _setUpvotePercent = (value) => {
     const { dispatch } = this.props;
```
feruzm, line 54: this.props won't work in functional component

src/components/upvote/container/upvoteContainer.js
```diff
@@ -22,17 +22,35 @@ import UpvoteView from '../view/upvoteView';
...
+  const {
+    content,
+    currentAccount,
+    fetchPost,
+    isLoggedIn,
+    isShowPayoutValue,
+    pinCode,
+    upvotePercent,
+    globalProps,
+    activeVotes = [],
```
feruzm, line 35: dispatch missing here

noumantahir:
I have updated the PR, can you please try running on your machine...or bundled version on android...

I do not want to overspend on this load posts fixes, need your review first, so far I can mark these changes as safe... but as soon I try my other pending fixes, it opens up a whole new set of breaches and introduce loop holes.

Let me know if I should work more on optimising this or do cyclic iterations, work on something else and come back to it later.

**Major change:**
- Introduced a new PostsList component with memo to avoid potential re-renders and optimised list cache:
still there are 2 more re-renders that are happening because of passed components (see screenshot), they go away if limit things to PostLists

Three pending fixes on my list that require post parser and posts fetch changes...
1. moving upvote calculator to post parser
2. fetching reblogs along with post/posts fetch
3. unlinking more of PostsList from PostContainer

1 and 2 does take away some processing efforts from FlatList, but it's affect is not well calculated, but do take away two more re-renders.

p.s. I have added some console statement for testing purposes... I will either remove them later to add a babel plugin to remove all logs while compiling a build

noumantahir:
Screenshot 2021-03-02 at 6 31 45 PM
Screenshot 2021-03-02 at 6 35 00 PM
Screenshot 2021-03-02 at 7 05 41 PM

feruzm:
I did some testing,
1. Opening any other feed and coming back to previous feed doesn't shows old list but new one. Example, click on any of the tags from feed, it should fetch you new feed according to that topic/tag and come back to previous feed, you will see posts from tag specific feed, so we should have feed data persistent for main feeds. Or going to profile page that fetches user's blog feed and come back to top/hot/new main feeds.
2. Voting on post from feed doesn't increase vote count, perhaps there is small bug there.
3. Let's remove fetching reblog info and reblog icon on feeds and move it to fetch when post is opened.

noumantahir:
Re 1: how about introducing a feed posts 2D sarray to persist data for each feed in main mainPossible issue: stored data in redux will increase substantiallyRe 2: will look into this bug, probably related to post refetchRe: 3: sure, if that doesn't affect app functions in feedsOn 3 Mar 2021 09:04, Feruz M <notifications@github.com> wrote:
I did some testing,
Opening any other feed and coming back to previous feed doesn't shows old list but new one. Example, click on any of the tags from feed, it should fetch you new feed according to that topic/tag and come back to previous feed, you will see posts from tag specific feed, so we should have feed data persistent for main feeds. Or going to profile page that fetches user's blog feed and come back to top/hot/new main feeds.Voting on post from feed doesn't increase vote count, perhaps there is small bug there.Let's remove fetching reblog info and reblog icon on feeds and move it to fetch when post is opened.

feruzm:
@noumantahir 
1. 2D array sounds good, but we should limit number of posts from each feed, let's say 5 posts from each feed and if user scrolls they will get rest of the list from api.
2. ok
3. Yes, that doesn't affect, we just don't show that information in front page, when user opens the post they can see reblog list, it might add processing into opening post page, which we can check for performance improvements next.

noumantahir:
This would fix both issues actually, making user wait every time a tab is changed, and data inconsistencies between screens.I do have an easy fix as well though, just resetting the cache everytime posts type changeOn 3 Mar 2021 09:23, Nouman Tahir <noumantahir@live.com> wrote:Re 1: how about introducing a feed posts 2D sarray to persist data for each feed in main mainPossible issue: stored data in redux will increase substantiallyRe 2: will look into this bug, probably related to post refetchRe: 3: sure, if that doesn't affect app functions in feedsOn 3 Mar 2021 09:04, Feruz M <notifications@github.com> wrote:
I did some testing,
Opening any other feed and coming back to previous feed doesn't shows old list but new one. Example, click on any of the tags from feed, it should fetch you new feed according to that topic/tag and come back to previous feed, you will see posts from tag specific feed, so we should have feed data persistent for main feeds. Or going to profile page that fetches user's blog feed and come back to top/hot/new main feeds.Voting on post from feed doesn't increase vote count, perhaps there is small bug there.Let's remove fetching reblog info and reblog icon on feeds and move it to fetch when post is opened.

noumantahir:
Got itOn 3 Mar 2021 09:47, Feruz M <notifications@github.com> wrote:
@noumantahir
2D array sounds good, but we should limit number of posts from each feed, let's say 5 posts from each feed and if user scrolls they will get rest of the list from api.okYes, that doesn't affect, we just don't show that information in front page, when user opens the post they can see reblog list, it might add processing into opening post page, which we can check for performance improvements next.

feruzm:
Right, present user cached feed but do fetch in background and notify user that there is new items on feed like this screenshot. `Fresh content available` or something similar popping up, user click would replace cache with new data that's already fetched **if new data is more than 5 posts, if less, just add them in top of cache and pop last items to keep 5 in cache**.
image

noumantahir:
Hmm..Will think around this nowOn 3 Mar 2021 09:58, Feruz M <notifications@github.com> wrote:
Right, present user cached feed but do fetch in background and notify user that there is new items on feed like this screenshot. Fresh content available or something similar popping up, user click would replace cache with new data that's already fetched if new data is more than 5 posts, if less, just add them in top of cache and pop last items to keep 5 in cache.

noumantahir:
@feruzm 
Issue 1: resolved by maintaining separate cache for feed screen and other screens using postsContainer... it is also step 1 in avoid posts reload every time user change tab on home screen.

noumantahir:
it have introduced a tiny glitch in profile screen, activity indicator appears twice... will look into this as well.

noumantahir:
@feruzm 
FIXED: Opening any other feed and coming back to previous feed doesn't shows old list but new one. Example, click on any of the tags from feed, it should fetch you new feed according to that topic/tag and come back to previous feed, you will see posts from tag specific feed, so we should have feed data persistent for main feeds. Or going to profile page that fetches user's blog feed and come back to top/hot/new main feeds. Refresh was probably need which you removed.
Remarks:
Fixed by adding a separate feed cache for posts other than feed screen, it has introduced a small UI glitch in profile screen though, but should be fixed as we progress. I am now relying on redux for posts source to FlatList, using it as base for custom caching to be added later

FIXED: Voting on post from feed doesn't increase vote count, perhaps there is small bug there.
Remarks:
Instead of fetching posts on upvote, I am faking it in UI..

SKIPPED: Let's remove fetching reblog info and reblog icon on feeds and move it to fetch when post is opened, it should decrease rerenders.
Remarks:
reblog count is actually being used on post list item, besides it is not contributing much in decreasing rending time. better if this count along with active_votes and image ratio to be returned by posts fetch api... it can make substantial difference.

noumantahir:
Not the best day for me...

**Some minor update from today**
added babel plugin to remove console logs from code
removed useless code form upvote view
put back 1 card post placeholder
added image caching

**What other I attempted and skipped for now**
- attempting to calculate thumb ratio right after fetch, still freezes app. probably the package itself is not well written for android
Solutions: 
1. we use pre-render function from FastImage or cache and transform image view in an optimised way. it uses native thread so should not affect performance
2. upgrade ImageSize package to use batch processing instead of processing single image, may improve performance, I can do it as a side thing contributing to open source.

- **Placeholder animation**: my initial idea was to use LayoutAnimation to achieve that, use native thread to render animation... but Lottie seems like a better option for looping animations on native thread without code.

- **Cached tab posts**: initially this seemed simple, however since we are using single component to manage both posts and tabs, as soon we cache posts, things start to misbehave, essentially because of too many states playing crucial part in maintaining right content on screen. 
`selectedFitlerValue`, `selectedFitlerIndex`, `startAuthor`, `startPermlink` just to name some. other challenges include, fetching new posts when needed and optimising cached posts in phone storage.
Solution:
I see this as a best place to implement useReducer and optimise redux insolvent in postsContainer. but given our current timeline for pushing an update on Friday, I believe there wont be enough time to test things around...

PR #1858, opened 2021-03-15
reduced max width available to left buttons of stickybar

noumantahir:
### What does this PR?
avoid trash icon pushed out of screen


feruzm (approved):
lgtm

PR #1859, opened 2021-03-16
Nt/tweaks

noumantahir:
### What does this PR?
- added keyboard aware view to keep snippets edit action panel above keybaord
- added mount flag in upvote container to avoid possible crash in release mode


feruzm (approved):
lgtm

PR #1861, opened 2021-03-17
Nt/tweaks

noumantahir:
### What does this PR?
- locked appcenter to 3.1.0
- using slice only when length greater than 5


feruzm (approved):
lgtm

PR #1865, opened 2021-03-18
Nt/tweaks

noumantahir:
### What does this PR?

- pin code modal fixes to achieve expected behaviour - check video
- themefied input and welcome modal

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/111686188-9729f980-884a-11eb-9294-6f0e1abd22d7.mp4

### DIFF
--- a/src/config/locales/en-US.json
+++ b/src/config/locales/en-US.json
@@ -323,7 +323,10 @@
     "enter_text": "Enter PIN to unlock",
     "set_new": "Set new PIN",
     "write_again": "Write again",
-    "forgot_text": "Oh, I forgot it..."
+    "forgot_text": "Oh, I forgot it...",
+    "pin_not_matched":"Pins do not match, Please try again.",
+    "attempts_postfix":"failed attempt(s)",
+    "message_reset_warning":"User data will be wiped on next failed attempt"
   },
   "alert": {
     "success": "Success!",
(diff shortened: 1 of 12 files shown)


noumantahir:
- handling mismatched pin while creating new pin
- added translation entries

noumantahir:
- fixed issue with login popup appearing upon login after reset,

noumantahir:
- showing pincode popup for re-logging in user that had pin set previously without reset.

noumantahir:
- patched up issue logging out on login using pin code

**Summary**
Issue was from loginContainer, the pinCode screen was being called only for verification while essentially user logging in with pinCode means the keys should also update according user set pin code.

Now I am also setting new keys in realm on login with pin code.

The reason I am calling this a patch is because this logout issue was taking too much time, when I figured the issue I just tried patch as cleanly as possible. but overall, the whole pinCodeContainer class can be cleaned to become something very simple.

noumantahir:
updated crypto.js to avoid UTF-8 malfunction issue during pin code verification

noumantahir:
Improved encryption and handling failure for pin code related decryptions.

src/config/locales/en-US.json
```diff
@@ -360,6 +360,7 @@
     "remove_alert": "Are you sure you want to remove?",
     "clear_alert": "Are you sure you want to clear?",
     "clear_user_alert": "Are you sure you want to clear all user data?",
+    "decrypt_fail_alert":"Bad previous app data, you will have relogin to reset app data",
```
feruzm, line 363: let's change string to this: `Corrupt app state, please relogin to reset state.`

noumantahir:
done

noumantahir:
- better handling decryption exception
- stamping new encryptions

feruzm (approved):
lgtm

PR #1872, opened 2021-03-22
Axios logs

noumantahir:
### What does this PR?
a goto solution for tracking axios api calls... reactotron is and will be a more robust solution but that involved some extra steps

does not include dhive logging yet.

### Screenshots/Video
Screenshot 2021-03-22 at 8 20 16 PM


noumantahir:
I had it mistakenly commit in local development branch... separated to axios-logs branch after... remote development (response, null, 2) branch wont be affected at all.

feruzm (approved):
lgtm

PR #1881, opened 2021-03-28
Nt/post draft master

### LINKED ISSUE #1877: locally saved drafts getting cleard
Nina (the user) she usually has issues with draft for some reason, draft save doesn't work for her, or she writes some text and goes to other screens and comes back and her draft is gone, etc. We should really check this draft flow. I know we save drafts to localstorage and backend as well, so there is combination when user is offline etc. cases, overall check on flow would be really good.

> Grrr! Updated to the new version of Ecency and found out the hard way the save draft button did not save all my work after over two hours inputting. Samsung Galaxy Note8 smartphone user.

1. Go to editor
2. Add some body, save draft
3. Go to any post, create and remove a replay draft
4. post draft gets cleared

**Expected behavior**
draft should only be cleared either on post publishing or user clear action

noumantahir:
### What does this PR?
tracks draft updates for simple editor and save drafts separately. if remote draft counter-part in more upto date, discards locally saved draft

### Issue number
fixes #1877 

### Screenshots/Video

https://user-images.githubusercontent.com/6298342/112744100-5fdbeb00-8fb6-11eb-87cf-8174049623fb.mov
disregard the slow performing packager, using my backup portable macbook.

@feruzm please review and let me know how we should improve this editor flow more...


feruzm:
Confusion here is that we have 2 different instance of same draft most cases, so let's update backend when user exit editor, app goes background or app changes page. Send latest update to backend draft, if never saved, save as new draft on backend.

So in essence, keep local and backend copy in sync, if backend respond fails for some reason or connection issue, keep local instance intact and retry again later when state changes again

noumantahir:
There are a few things to consider.

If save draft to DB on exist, what do user see in editor when coming back to editor using mid tab.
- in case we show latest db saved draft, it would raise an issue, what if user intend to create a separate post.
- if we show local copy independent of db saved draft, in this case on exit routine will be save a new instance of draft possibly an updated copy.

The issue do still exist in previous app as well though, but updates so far tend to resolve the cleared local cache on exit.

feruzm:
@noumantahir does this include #1880 also or that one should be separately merged?

noumantahir:
created new PR instead

PR #1885, opened 2021-03-30
removed timeout - using post body loading flag to render footer content

### LINKED ISSUE #1883: Post details load delay is sometimes more than 200 ms
**Describe the bug**
"on my device, I don't see any problems on feeds, only post opening is slower compare to previous production release, I suppose animation timeout causing it, because on iOS we never had performance issues only on Android sides."

noumantahir:
### What does this PR?
- Instead of using timeout I have reverted code to using old way of render placeholder
- to avoid momentary glitch of comments tab appearing before post body is display, I am now rendering all scrollable footer content after post body is done rendering content.

### Tradeoff
- like before, the post with large body will still show a momentary gap between placeholder animation going away and post body being rendered... however the difference now is the removal of comments glitch which should cater for the UX issue from older versions.

### Issue number
fixes #1883

### DIFF
--- a/src/components/postElements/body/view/postBodyView.js
+++ b/src/components/postElements/body/view/postBodyView.js
@@ -31,6 +31,7 @@ const PostBody = ({
   handleOnUserPress,
   handleOnPostPress,
   dispatch,
+  onLoadEnd,
 }) => {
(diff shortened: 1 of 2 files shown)


feruzm (approved):
lgtm

PR #1896, opened 2021-04-05
fixed notification data getting ignore after refresh

noumantahir:
### What does this PR?
- fixed notification data getting ignore after refresh


feruzm (approved):
lgtm

PR #1899, opened 2021-04-05
added image delete confirmation alert

noumantahir:
### What does this PR?
User is now asked for confirmation whether to proceed with image delete or cancel

### Screenshots/Video
Screenshot 2021-04-06 at 1 38 21 AM

### DIFF
--- a/src/config/locales/en-US.json
+++ b/src/config/locales/en-US.json
@@ -337,7 +337,11 @@
     "title_remove_confirmation":"Delete image",
     "btn_add":"Image",
     "message_failed":"Failed to upload image",
-    "message_remove_confirmation":"Are you sure you want to delete this image?"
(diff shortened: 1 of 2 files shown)


src/config/locales/en-US.json
```diff
@@ -337,7 +337,11 @@
     "title_remove_confirmation":"Delete image",
     "btn_add":"Image",
     "message_failed":"Failed to upload image",
-    "message_remove_confirmation":"Are you sure you want to delete this image?"
+    "message_remove_confirmation":"Are you sure you want to delete this image?",
+    "alert_remove_title":"Confirm Delete",
+    "alert_remove_body":"Are you sure you wish to delete this image?",
```
feruzm, line 342: `message_remove_confirmation` is same, we could use that, which might be already translated into different languages.

PR #1902, opened 2021-04-07
Nt/native filters

### LINKED ISSUE #1888: Native filter tabs
Current filter system is used more like button to shift state causing posts feed to re-render posts list items. This causes a momentary lag in app performance.

**Expected behavior**
The transition from one tab to another should be instantaneous. just like bottom navigation tabs only in this case it should change posts filter.

**Additional context**
The idea is to replace button based filter tab with actual native tabs with built-in cache functionality. 

Ideally the new system will use most of the same codebase as available postsContainer and sub components

**Checklist**

- [x]  promoted posts
- [x]  user switching hooks - (buggy when switching user while in non feed tab)
- [x]  double tap to scroll up
- [x]  load posts from storage on startup
- [x]  handling logout
- [x]  Make hide Images button functional
- [x]  scheduled new posts check
- [x]  Fix memory leak on component unmount
- [x]  fix missing keys issues
- [x]  traverse on scroll event back to component parent
- [x]  post tabbedPosts to profile screen
- [x]  port tabbedPosts to community pages
- [x]  port tabbedPosts to tag results pages

noumantahir:
!!!Strictly a work in Progress, yet please feel free to test

### What does this PR?
Upgraded main feed screen to use native filters, it is still a work in progress, but the feed screen is now useable with most of primary functions like, 

1. pagination
2. refresh
3. logged out view
4. empty list view for friends and communities tabs

Pending Items
1. promoted posts
2. user switching hooks
3. double tap to scroll up
4. new posts check

fixes #1888 

Video
https://youtu.be/ZDgcxsvxw-k


noumantahir:
@feruzm you may review this now....

I could not find a fix for nested ScrollableTabView unresponsive tab change while scroll issue.

Suggestion:
We can use the TabbedPosts to accept external content overrides and get rid of nested tabs at all..

noumantahir:
fixed scrolling issue by discarded nested tabview in favour for stacked tabbar

noumantahir:
Editable tags are also merged with native filter since they work in collaboration with each each other.

fixes #1913

feruzm:
@noumantahir Could you update thread with what's pending and what needs to be tested more? We could merge and do more tests on device if nothing major is pending.

noumantahir:
@feruzm we can push it device for testing... there is nothing pending on the filters side.... 

The second major part we discussed related to postsContainer was conversion on PostCard to PureComponent, but I can safely postpone that as app has become substantially performance even without that. It is however a must do upgrade as well.

feruzm:
PureComponent we use in some parts, if done properly I do agree it can give good performance boost, potentially getting rid of unnecessary re-renders. Let's do it in separate PR.

feruzm (approved):
lgtm, let's do some test with actual devices 🎉

PR #1917, opened 2021-04-13
Nt/post parser fix

noumantahir:
### What does this PR?
using catchPostImage instead of json_meta for cover image


feruzm (approved):
lgtm

PR #1925, opened 2021-04-15
Nt/tweaks

noumantahir:
### What does this PR?
- fixed continous refreshing issue
- fixed latest post check on startup
- fixed crash on back button press after reply
- updated search input in searchResultScreen


feruzm (approved):
lgtm

PR #1930, opened 2021-04-23
Tweaks

noumantahir:
### What does this PR?
- welcome screen touchable styling
- fixes image sharing to editor (testing required, simulator causing issues)
- attempted to fix possible tag issue reported by ecency user yayogerado

### DIFF
--- a/src/screens/editor/container/editorContainer.js
+++ b/src/screens/editor/container/editorContainer.js
@@ -103,14 +103,14 @@ class EditorContainer extends Component {
 
         upload.forEach((el) => {
           if (el.filePath && el.fileName) {
-            this.setState({ isUploading: true });
+            // this.setState({ isUploading: true });
             const _media = {
               path: el.filePath,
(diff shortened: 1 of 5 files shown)


src/screens/editor/container/editorContainer.js
```diff
@@ -103,14 +103,14 @@ class EditorContainer extends Component {
 
         upload.forEach((el) => {
           if (el.filePath && el.fileName) {
-            this.setState({ isUploading: true });
+            // this.setState({ isUploading: true });
```
feruzm, line 106: any reason why we commenting this? it should show progress/spinner while upload is happening, doesn't this break that?

noumantahir: spinner still kicks in from `_uploadImage` function. Also, for logged out user it was causing spinner to show indefinitely

feruzm (approved):
lgtm

PR #1943, opened 2021-05-09
Pin modal disappearing fix

noumantahir:
1. introduced version number parser
2. fixed modals disappearing on startup when version changes

### Trade off
If version changes, the welcome modal appears first before pin code modal... this was the easiest hack, otherwise the logic was ending up becoming dirty for not obvious reason. the source of bug bug two modal appears triggering at almost the same instance as there was a slight delay in redux shooting modal required update

### Reference
https://trello.com/c/FI5wUDVC/38-bug-pin-code-enabling-and-welcomescreen-on-wont-ask-pin-code-for-some-reason-also-welcome-screen-auto-closes

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/117557609-23f47500-b08e-11eb-86fd-658f89e54c9d.mov

### DIFF
--- a/src/realm/realm.js
+++ b/src/realm/realm.js
@@ -1,5 +1,6 @@
 import sha256 from 'crypto-js/sha256';
 import AsyncStorage from '@react-native-community/async-storage';
+import parseVersionNumber from '../utils/parseVersionNumber';
 
 // CONSTANTS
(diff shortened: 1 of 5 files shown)


feruzm (approved):
lgtm

PR #1944, opened 2021-05-09
Youtube iFrame

noumantahir:
integrated youtube iframe as action sheet. It does seem to work well across all devices.. 

Though on emulator it's only plays audio but that should not affect actual device as it is a know issue 
https://github.com/LonelyCpp/react-native-youtube-iframe/issues/29

### Known Issue
On android, when video is in fullscreen and user change device orientation, the app triggers a reset and gets stuck on loading screen. issue can be seen in video as well.
Any solution suggestion from your experience?

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/117560096-44c7c500-b0a4-11eb-8fbb-89fdd77a3a63.mov

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,6 +105,7 @@
     "react-native-version-number": "^0.3.5",
     "react-native-webview": "^11.2.1",
     "react-native-youtube": "^2.0.1",
+    "react-native-youtube-iframe": "^2.0.1",
     "react-navigation": "^4.0.10",
     "react-navigation-drawer": "^2.3.3",
     "react-navigation-redux-helpers": "^4.0.1",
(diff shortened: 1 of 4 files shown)


feruzm:
@noumantahir one solution is to force disable landscape mode, I believe we did that maybe something we can do on video landscape also? Another solution is to find out what is the reason for this splash stuck, because there was another instance when this occur

noumantahir:
@feruzm 

1. full-screen video is essentially separate entity, I have not found a solution to lock orientation so far

2. app getting suck in loading animation is a separate issue that need deep checking, but even that fix this issue, as target is to avoid app reset

feruzm (approved):
let's test this and we can fix known issues in separate PR.

PR #1946, opened 2021-05-11
auto install pods when yarn install is done

noumantahir:
runs `pod install` every time packages are updated using `yarn`. simple script to take away hassle of doing pod install every time


feruzm (approved):
maybe needed to auto exit folder after `pod install`? Let's see if it won't break CI

PR #1949, opened 2021-05-19
Nt/custom tabs

noumantahir:
### What does this PR?
- Support for tab customisation in profile and community screens
- own profile and others tab setting is dealt separately
- Introduced profile component children commentsTabContent that handles fetch for replies and comments tab.

More of comments separate component
>   handling comments and replies was a bit of challenge, there were unnecessary calls and adding support within profileContainer might have made code more dirty. 
The new component use existing Comments component for view rendering however, I have unlinked the fetchPost callback that used to fetch comments as it was causing uncontrolled calls and serve no benefit as even if new comment is made, it does not appear in api calls for some time.
The 
These extra calls existed before as well and there are many other such calls in app, I identified them using reactotron, perhaps we can gradually clean these calls overtime to improve performance where it's due.

> The children and primary container structure is something I suggest to upgrade whole app to follow. the concept is, instead of having container and view sub directories we can have container and children, the children will container all components that are only used within that container and not reused anywhere else in app.

### Videos

-- Comments Replies response
https://user-images.githubusercontent.com/6298342/118784546-8d079400-b8a9-11eb-8750-22b01de5af12.mov
*Ignore the abrupt dialog appearing, my mouse is malfunctioning today

-- Customising tabs
https://user-images.githubusercontent.com/6298342/118784519-8711b300-b8a9-11eb-9557-5c8c0801148b.mov


feruzm (approved):
lgtm 🚀

PR #1952, opened 2021-05-22
NT - Own Profile Fix

noumantahir:
Allowing user to access profile from post header

instead of opening tab, the profile opens an a regular screen as otherwise tab opening can lead to breaking experience is access from deeper stage of navigation..

let me know your thoughts.


feruzm (approved):
lgtm

PR #1957, opened 2021-05-25
added support for heic image links in meta

noumantahir:
Backtracking image issue turned out to be indeed heic not being read as a legit image source. Have updated the regex to support heic. It's untested at the moment, waiting for me iPhone, 15 - 30 minutes.


feruzm (approved):

feruzm:
Let’s merge and double test with TestFlight also

PR #1959, opened 2021-05-25
NT - DRAFT/SCHEDULED POST BUG

noumantahir:
fixed issues with drafts screen

1. picking heic image
2. smartly rendering title, summary and thumbnail based on availability
3. renamed postListItem to draftListItem to avoid confusings

Screenshot 2021-05-25 at 4 08 55 PM


src/utils/image.js
```diff
@@ -64,7 +64,7 @@ export const catchEntryImage = (entry, width = 0, height = 0, format = 'match')
 };
 
 export const catchDraftImage = (body, format = 'match', thumbnail = false) => {
-  const imgRegex = /(https?:\/\/.*\.(?:tiff?|jpe?g|gif|png|svg|ico|PNG|GIF|JPG))/g;
```
feruzm, line 67: Is regex case insensitive usually?

src/screens/drafts/screen/draftsScreen.js
```diff
@@ -10,7 +10,7 @@ import { catchDraftImage } from '../../../utils/image';
 import { getFormatedCreatedDate } from '../../../utils/time';
 
 // Components
-import { BasicHeader, TabBar, PostListItem, PostCardPlaceHolder } from '../../../components';
```
feruzm, line 13: We are not using PostListItem anywhere else now that warrant name change to draftlistitem?

feruzm (approved):
lgtm

PR #1961, opened 2021-05-26
Nt/foreground notification

noumantahir:
foreground notification appears for replies and reblogs..... it is untested before making PR, I ran out of credits to make comment

Screenshot 2021-05-26 at 8 48 53 PM


feruzm (approved):
lgtm

PR #1962, opened 2021-05-29
Nt/refres fcm token

noumantahir:
### What does this PR?
Keeping local notification settings and remote settings plus token in sync by updating backend on startup

### DIFF
--- a/src/screens/application/container/applicationContainer.js
+++ b/src/screens/application/container/applicationContainer.js
@@ -628,7 +628,7 @@ class ApplicationContainer extends Component {
   };
 
   _getSettings = async () => {
-    const { dispatch } = this.props;
+    const { dispatch, otherAccounts } = this.props;
 
     //reset certain properties
     dispatch(hideActionModal());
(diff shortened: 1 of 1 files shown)


src/screens/application/container/applicationContainer.js
```diff
@@ -854,7 +882,6 @@ export default connect(
     // Application
     isDarkTheme: state.application.isDarkTheme,
     selectedLanguage: state.application.language,
-    notificationSettings: state.application.isNotificationOpen,
```
feruzm, line 857: Settings and Login screens use this I believe, wouldn't that get broken if removed?

noumantahir: I doubt that, it is not being used in this reference

noumantahir: giving it a second look just in case

noumantahir: I am pretty confident it won't break anything, check the screenshot fro details
Screenshot 2021-05-29 at 12 51 23 PM

feruzm (approved):
great, lgtm 👍

PR #1982, opened 2021-06-23
Nt/api migration

noumantahir:
### What does this PR?

Setup a basic structure for using local created access token. To avoid repeated code lines I am using ecencyApi interceptor to inject upto date accessToken from redux state instance.

Have tested the get drafts method using this and seems work well.

### Update 1:

1. Fixed issue with accessToken getting overwritten by realm data.
2. refreshing access token even if auth type is mastKey
3. Added migration script that check and updates accessToken for masterKey logins at 3 places

3.1. Startup
3.2. Account Switch
3.3. Logout User Switch

### Update 2:
Migrated drafts api to use ecencyApi 
Using typescript and async/await
For testing make sure to update `.env` to use `ECENCY_BACKEND_API=https://ecency.com/` it was using `/api/` before it is not valid.

### Update 3:
All available and being used endpoints in app are migrated... below is a list of pending ecency.ts endpoint that are not available from ecencyApi yet...

```
/report
/market-data/currency-rate/
/rgstrmbldvc/ (setPushToken)
/purchase-order
/post-reblogs/

```
### Known Issues
_MOVED TO SEPARATE TASK:_ Second issue is special case that might occur, yet to verify.. When access token is refreshed, the routine does not update the redux data but only the realm that means redux will not have latest tokens uptill next app session that can lead to potential api failures


src/providers/hive/auth.js
```diff
@@ -62,6 +64,19 @@ export const login = async (username, password, isPinCodeOpen) => {
     }
   });
 
+  // Prepare hivesigner code
+  const signer = (message) => {
+    const hash = cryptoUtils.sha256(message);
+    return new Promise((resolve) => {
+      const key = privateKeys.activeKey;
```
feruzm, line 71: message can be signed with `ownerKey || activeKey || postingKey`

noumantahir:
The PR can now be tested and merged with development...

src/utils/b64.ts
```diff
@@ -0,0 +1,13 @@
...
+    '+': '-',
+    '-': '+',
+    '=': '.',
+    '.': '=',
+  };
+  
+  export const b64uEnc = (str) =>
+    Buffer.from(str)
+      .toString('base64')
+      .replace(/(\+|\/|=)/g, (m) => b64uLookup[m]);
```
feruzm, line 13: new line missing at the end

src/screens/notification/container/notificationContainer.js
```diff
@@ -29,21 +29,20 @@ class NotificationContainer extends Component {
...
 
-    if (username && isConnected) {
-      this._getAvtivities(username);
+    if (isConnected) {
+      this._getAvtivities();
     }
   }
 
-  _getAvtivities = (user, type = null, loadMore = false) => {
+  _getAvtivities = (type = null, loadMore = false) => {
```
feruzm, line 39: this could be renamed perhaps to `_getActivities`?!

feruzm (approved):
lgtm, let's do some tests on device

PR #1983, opened 2021-07-05
NT - Local Vote Duplication

noumantahir:
The duplicated was being caused by comment not having any associated post-id, now I am using combination of author and permlink as unique postId.

https://user-images.githubusercontent.com/6298342/124440770-6a1b3b80-dd94-11eb-9920-fad9502c15fa.mov


feruzm (approved):
great, author/permlink combination always unique

PR #1984, opened 2021-07-06
NT - Functional Upvote View

noumantahir:
Converted upvote view to functional component
Avoiding using UNSAFE methods
Improved sliderValue calculator


feruzm (approved):
lgtm

PR #1986, opened 2021-07-06
Nt/upvote bug

noumantahir:
### What does this PR?

Updated estimate amount calculator to be more redundant.
Replaced esteemapp popover package wiht one from doomsower as apparently it seems more upto date. Though it still uses UNSAFE method but let's hope it fixes the android popover crash else we can attempt to upgrade popover ourselve, actually we should but let's do it afterwards.

Screenshot 2021-07-06 at 6 45 51 PM


feruzm (approved):
lgtm

PR #1987, opened 2021-07-07
Nt/clearing draft

noumantahir:
Updated post submit flow.

1. While publishing a post, on successful response the draft is removed from ecency server if available
2. Similarly while scheduling, the associated draft is deleted on success
3. The recent draft logic is updated as it was causing issues after publish post draft delete
4. Added isPublishing flag to avoid possible publish duplication, though it may not be the reason for the problem shared by user, but should add some extra redundancy.


feruzm (approved):
lgtm

PR #1988, opened 2021-07-09
Nt/bug fixes

noumantahir:
Screenshot 2021-07-09 at 6 57 39 PM

### Pending fixes...

**Updating Add To Bookmarks title for bookmarked post.**
Remarks: no checkBookmark endpoint available like available for favourites, will have to maintain local cache of all bookmarks, may not be the most clean solution, that is why not prioritising. 

**Image picker malfunctioning on MI Redmi Note 10**
Tested on different android devices, seems to work fine but not working on my device, it seems like a known issue, added a comment on a relevant issues thread.

**Upvote to Downvote change does not work as expected visually**
Yet to dig to see if that is a local cache issue or something else.


feruzm (approved):
lgtm

PR #1989, opened 2021-07-09
Nt/bug fixes

noumantahir:
Updates:
fixed issue with upvote downvote local cache swap not updating ui
fixed issue with comment parser total value

Screenshot 2021-07-09 at 9 32 35 PM


feruzm (approved):
lgtm

PR #1990, opened 2021-07-12
Nt/fixes

noumantahir:
Fixed issues with notifications tab.

1. mark all read is working now
2. fixed pagination fetch issue
3. fixed pagination ui load glitch

Other issue: the tab shift feels really laggy in notifications, we should make it fluid

https://user-images.githubusercontent.com/6298342/125285571-38761780-e334-11eb-85bf-65582a9f5ac2.mov

Updated profile edit button textual button
Screenshot 2021-07-12 at 5 10 40 PM


feruzm (approved):
lgtm

PR #1991, opened 2021-07-12
NT Token Refresh

noumantahir:
Refreshing access token on every new session, account switch and pin …

The concept is, wherever current account is set using UpdateCurrentAccount action access token is refreshed and updated.

After this release or perhaps at some suitable time, we should restructure the application container and state initialization to improve startup performance.


feruzm (approved):
lgtm, let's test on device

PR #1992, opened 2021-07-12
NT Android App Links

noumantahir:
updated intent filters to support custom app link paths
rest of paths like faq and proposals open up in browser just like we did in iOS

Tested links:
Screenshot 2021-07-13 at 12 39 24 AM


feruzm (approved):
lgtm

PR #1994, opened 2021-07-13
NT Vote Estimation

noumantahir:
updated estimate vote value calculation recipe to match with Hive recipe

Screenshot 2021-07-13 at 1 31 49 PM
Screenshot 2021-07-13 at 1 37 22 PM


feruzm (approved):
Lgtm

src/utils/conversions.ts
```diff
@@ -0,0 +1,17 @@
...
+};
+
+export const vestsToRshares = (vests:number, votingPower:number, weight:number) => {
+  if (!vests || !votingPower || !weight) {
+    return 0;
+  }
+
+  const finalVest = vests * 1e6;
+  const power = (votingPower * weight / 10000) / 50
+  return (power * finalVest / 1000)
```
feruzm, line 16: @noumantahir here one 0 missing, no? used to be `/1e4`

noumantahir: Will compare shortly

noumantahir: I double checked, only thing I could find varying is weight, as I am used weight calculation from previous formula as not information given in their documentation.

`  const weight = sliderValue * 10000;`

PR #2007, opened 2021-07-24
NT - Community Tag

noumantahir:
Put back community tag in feed, kept removed from post details

### Screenshots/Video
Screenshot 2021-07-24 at 8 33 29 PM

**Suggestion**
Regarding community tag of post details...
If we want to show community tag in post header, may be we should remove community tag from list of tags below, this way we will not have double entries while creating a distinction between regular tags and community.
Screenshot 2021-07-24 at 8 33 51 PM

### DIFF
--- a/src/components/postView/view/postDisplayView.js
+++ b/src/components/postView/view/postDisplayView.js
@@ -221,7 +221,6 @@ const PostDisplayView = ({
                 name={author || post.author}
                 currentAccountUsername={name}
                 reputation={post.author_reputation}
-                content={post}
                 size={36}
               />
               <PostBody body={post.body} onLoadEnd={_handleOnPostBodyLoad} />
--- a/src/components/postElements/headerDescription/view/postHeaderDescription.js
+++ b/src/components/postElements/headerDescription/view/postHeaderDescription.js
@@ -89,6 +89,16 @@ class PostHeaderDescription extends PureComponent {
             </View>
           </View>
           <View style={styles.rightContainer}>
+            {content && (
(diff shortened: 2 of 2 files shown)


src/components/postView/view/postDisplayView.js
```diff
@@ -221,7 +221,6 @@ const PostDisplayView = ({
                 name={author || post.author}
                 currentAccountUsername={name}
                 reputation={post.author_reputation}
-                content={post}
```
feruzm, line 224: @noumantahir Are you sure we don't need this props anymore?

noumantahir:
Yes, it was only used to render the community tag in profile header.... Removing it from postDisplay removed only the tag.I did double check for its usecaseOn 24 Jul 2021 11:40 pm, Feruz M ***@***.***> wrote:
@feruzm commented on this pull request.



In src/components/postView/view/postDisplayView.js:
> @@ -221,7 +221,6 @@ const PostDisplayView = ({
                 name={author || post.author}
                 currentAccountUsername={name}
                 reputation={post.author_reputation}
-                content={post}

@noumantahir Are you sure we don't need this props anymore?

src/components/postView/view/postDisplayView.js
noumantahir: @feruzm 
Screenshot 2021-07-24 at 8 59 48 PM

feruzm: ok

feruzm (approved):

PR #2008, opened 2021-07-24
NT - Meta Fix

noumantahir:
parsing json_metadata as object for comments

### Screenshots/Video
BEFORE
Screenshot 2021-07-24 at 8 56 03 PM

AFTER
Screenshot 2021-07-24 at 8 59 48 PM

### DIFF
--- a/src/utils/postParser.js
+++ b/src/utils/postParser.js
@@ -86,7 +86,6 @@ export const parseComments = async (comments) => {
     if (typeof comment.json_metadata === 'string' || comment.json_metadata instanceof String) {
       try {
         comment.json_metadata = JSON.parse(comment.json_metadata);
-        comment.json_metadata.tags = ['hive-123123'];
       } catch (error) {
         comment.json_metadata = {};
       }
--- a/src/utils/postParser.js
+++ b/src/utils/postParser.js
@@ -82,6 +82,15 @@ export const parseComments = async (comments) => {
     comment.markdownBody = get(comment, 'body');
     comment.body = renderPostBody(comment, true, webp);
 
+    //parse json meta;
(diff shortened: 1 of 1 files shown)


src/utils/postParser.js
```diff
@@ -82,6 +82,15 @@ export const parseComments = async (comments) => {
     comment.markdownBody = get(comment, 'body');
     comment.body = renderPostBody(comment, true, webp);
 
+    //parse json meta;
+    if (typeof comment.json_metadata === 'string' || comment.json_metadata instanceof String) {
+      try {
+        comment.json_metadata = JSON.parse(comment.json_metadata);
+        comment.json_metadata.tags = ['hive-123123'];
```
feruzm, line 89: Please check this. IF tags is empty we set default tag, not always. And it has to be `['hive-125125']` (Ecency community) not 123123...

noumantahir:
Good catch and my bad, I had it set for testing, should not be thereOn 24 Jul 2021 11:45 pm, Feruz M ***@***.***> wrote:
@feruzm commented on this pull request.



In src/utils/postParser.js:
> @@ -82,6 +82,15 @@ export const parseComments = async (comments) => {
     comment.markdownBody = get(comment, 'body');
     comment.body = renderPostBody(comment, true, webp);
 
+    //parse json meta;
+    if (typeof comment.json_metadata === 'string' || comment.json_metadata instanceof String) {
+      try {
+        comment.json_metadata = JSON.parse(comment.json_metadata);
+        comment.json_metadata.tags = ['hive-123123'];

Please check this. IF tags is empty we set default tag, not always. And it has to be ['hive-125125'] (Ecency community) not 123123...

feruzm (approved):

PR #2021, opened 2021-08-04
Nt - Notification Pagination Fix

noumantahir:
discard nested flatlist in favour of sectioned list.

Can you please try it out first, waiting for your test results, I only simulated the problem locally since I was not experiencing it with my accounts.


feruzm (approved):
let's try on device

PR #2022, opened 2021-08-04
Nt/iap upgrade

noumantahir:
Apps build successfully with updated package. Overall experience require testing


feruzm (approved):
let's test on device

PR #2044, opened 2021-08-17
Nt - Scrolling Comment

noumantahir:
setting post display scroll to have min height equals window

android web view do not support scrollEnabled property and when scroll viewport is less than scrollable height, the webview was letting the content to be scroll, this was happening for both comment and body.

now, since height will always be scrollable, content of webview is set to not scrollable.

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/129716394-ad683887-7a69-48c3-be4f-d5d1c5d0ace4.mov


feruzm (approved):
let's test on device.

PR #2047, opened 2021-08-18
Nt - JCenter Removal

noumantahir:
Shifted repo to mavenCentral with exception of few selective group
Remove when replacement for selective groups is available the jcenter will be accessible as read-only indefinitely so we should be good for now with these selective packages as they are not available off of jCenter

Selective groups:
            includeGroup("com.facebook.yoga")
            includeGroup("com.facebook.fbjni")
            includeGroup("com.henninghall.android")
            includeGroup("org.matomo.sdk")

However since they are now read-only means no updates will be received by jCenter... we should check back at some point in future if we can get rid of these groups as well.

App does build successfully in my tests with this setup

Let me know your thoughts.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -23,8 +23,10 @@ buildscript {
         firebaseVersion = "17.3.4"
     }
     repositories {
+        mavenCentral()
         google()
-        jcenter()
+
+        
     }
     dependencies {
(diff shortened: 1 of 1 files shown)


noumantahir:
updated

feruzm (approved):
lgtm

PR #2084, opened 2021-10-02
adjusting json_metadata tags to not be string

noumantahir:
### What does this PR?
It seems that some hive app or some logic is setting tags to be a string if there is just one tag...that was causing the crash of frontend, though there was also one bug of array check not working as expected but  now the issue has been fixed in the root or postParser.

### Screenshots/Video
NOTE: notification was rerouted to troublemaking comment for testing purpose, the test code is removed from PR.
https://user-images.githubusercontent.com/6298342/135722885-3e140624-b9f7-4708-830b-aead0cc5f0cd.mov


feruzm (approved):
Lgtm

PR #2085, opened 2021-10-02
Nt/options modal

noumantahir:
- Introduced options model that better handle the onPress callback
- Migrated components to options model that were relying on side effects to trigger ActionSheet

To summarise it resolves the missing option menu and wrong state referencing issues that raises using OptionsModel. My first attempt was to fork react-native-actionsheet but that consumed too much without even getting close to solution, the OptionsModal serves as a good replacement, though it uses actionsheet but fixes issue, another benefit is we can easily expand this to out own custom options model in future is ever need.

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/135732572-1318ba31-6ccb-4d3f-90e1-14c59487d84a.mov


feruzm (approved):
Lgtm, let’s test it in actions

PR #2086, opened 2021-10-03
form validity fix

noumantahir:
total of 10 tags are allowed but publish button was being disabled at 9 because of wrong comparison

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/135743713-8ea21d25-8905-487d-8a89-5729576c2be1.mov


feruzm (approved):
Oh good find 😮

PR #2097, opened 2021-11-26
Nt/follow btn

noumantahir:
### What does this PR?
1. disabling follow button while action in loading state for both profile screen and modal
2. avoid fetching profile after follow unfollow action as previous transaction value overwrite the new, and double profile fetch do not always yield the right result, instead rely on cached state.

### Current issue
in video will see a state where modal shows user is following while profile show user is not following, that is because of transaction delay and the fact cache is only limited to state of profile screen.
Ideally user may not be annoyed as status should update in a second or two but if we want fix that we can maintain follow state at redux level, but for now major bug is resolved. Let me know your thoughts.

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/143588916-467545d2-5bf4-4c96-8434-d947f5aeadbd.mov

### DIFF
--- a/src/components/organisms/quickProfileModal/children/actionPanel.tsx
+++ b/src/components/organisms/quickProfileModal/children/actionPanel.tsx
@@ -1,14 +1,14 @@
 import * as React from 'react';
-import { Text, View, StyleSheet } from 'react-native';
+import { View } from 'react-native';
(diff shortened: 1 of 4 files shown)


feruzm (approved):
lgtm

PR #2099, opened 2021-11-30
Nt/max payout

noumantahir:
### What does this PR?
- max payout overrides post card value is pending payment exceeds max
- added entry for max accepted payout in popover is limit is exceeded

### Screenshots/Video
Screenshot 2021-11-30 at 2 40 41 PM

Update:
- fixed currency of popover
Screenshot 2021-12-02 at 9 27 44 PM


feruzm (approved):
lgtm

PR #2100, opened 2021-12-03
Nt/feed refresh

noumantahir:
### What does this PR?
fixes feed refresh issue if primary posts dataset is same, now upon refresh or launch, app discards previous state even feed data contains same posts.


feruzm (approved):
Let’s do some tests

PR #2103, opened 2021-12-16
Nt/post duplication fix

noumantahir:
### What does this PR?
Editor navigators now use discrete keys to differentiate between editor.. 

The reason why post was being duplicated is because when navigating for editing right after making a post, navigator was going back to previous editor with publish perspective. Now is initialise a new editor


feruzm (approved):
lgtm

PR #2104, opened 2021-12-16
fixed search tag not changing bug

aliseyalvi:
### What does this PR?
Fixed Bug: tapping post tag within tag search does not update search tag.
Bug was causing to not change the tag value when tapping on tag. Fix was to change the state of selected tag when tag value changes on component mount.

### Screenshots/Video
https://www.loom.com/share/77459df6d0a44fceaa541192549decdc


noumantahir (commented):
seems to worked well, tested on iOS sim

feruzm (approved):

PR #2105, opened 2021-12-16
using fast image with grey background for img html renderer

noumantahir:
### What does this PR?
Post images now appear with a grey placeholder instead of having an empty space filled with an image out of no where.. the grey placeholder also works as an indication for user that data is being loaded.

Since this change is targeting an essential component of app, we should test it very well.

### Screenshots/Video

BEFORE
https://user-images.githubusercontent.com/6298342/146389094-7efbad1a-c8de-4356-b74f-bf81e5afa002.mov

AFTER
https://user-images.githubusercontent.com/6298342/146389202-26137eb2-16e1-4966-b9b3-d1d59ced5a48.mov


feruzm (approved):
lgtm

PR #2106, opened 2021-12-16
opening witness link externally as done on ecency web version

noumantahir:
### What does this PR?

the witness link was being ignore previously for some reason, I have enabled it to open witness link externally.

Let me know if it was being ignore on purpose or by mistake, in any case, it was giving a bad UX so now at-least link open externally just like website.

Let me know if we should handle in some other way.

### Screenshots/Video
BEFORE/AFTER COMPARISON

https://user-images.githubusercontent.com/6298342/146431236-07119be2-362b-4d03-ac62-4504136a2662.mov


feruzm (approved):
lgtm

PR #2109, opened 2021-12-19
image size fix

noumantahir:
### What does this PR?
using external image ratio calculator, apparently on android FastImage was returning wrong image sizes, now we are using getSize external method from Image to set image ratio... 

Though the content rendering have become more smooth yet now we are making double the amount of network calls for single image under the hood. We can further try maintaining a long cache of image ratios to avoid that, yet we should first test the current implementation and discuss.

### Screenshots/Video

https://user-images.githubusercontent.com/6298342/146679776-b74f900e-9262-4b33-9111-184dfedff5d0.mov

### DIFF
--- a/src/components/autoHeightImage/autoHeightImage.tsx
+++ b/src/components/autoHeightImage/autoHeightImage.tsx
@@ -1,0 +1,60 @@
+import React, { useEffect, useState } from "react";
+import { Image } from "react-native";
+import EStyleSheet from "react-native-extended-stylesheet";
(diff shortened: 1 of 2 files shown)


feruzm:
@noumantahir `we are making double the amount of network call` is fine, but we can make first call with very small size of image, say 10x10 and get ratio from that image and 2nd call will be proper image size. Since we are making 2 calls we might also consider showing first image with blurry transition to full image.

feruzm (approved):

PR #2111, opened 2021-12-21
added meta in addDraft and updateDraft

aliseyalvi:
### What does this PR?
Added meta in addDraft api and updateDraft api endpoint. Uses extractMetaData utility from editor to extract the metaData from body and sent it to drafts-add and drafts-update endpoint.
### Issue number
Trello: Drafts API now has meta field as well, let's utilize that

### Screenshots/Video
https://www.loom.com/share/378cdf696be54927a9176e9e34f80fa0

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -10,13 +10,13 @@ PODS:
     - AppCenter/Crashes (~> 4.0)
     - AppCenterReactNativeShared (~> 4.0)
     - React-Core
-  - AppCenter/Analytics (4.2.0):
+  - AppCenter/Analytics (4.3.0):
     - AppCenter/Core
-  - AppCenter/Core (4.2.0)
-  - AppCenter/Crashes (4.2.0):
+  - AppCenter/Core (4.3.0)
+  - AppCenter/Crashes (4.3.0):
     - AppCenter/Core
-  - AppCenterReactNativeShared (4.2.0):
-    - AppCenter/Core (= 4.2.0)
+  - AppCenterReactNativeShared (4.3.0):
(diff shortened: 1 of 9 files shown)


noumantahir:
@aliseyalvi saving part is working well, but on loading a draft the thumbnail is not being set properly....check screenshots below, also updating your test draft....

Screenshot 2021-12-22 at 3 54 09 PM
Screenshot 2021-12-22 at 3 54 59 PM
Screenshot 2021-12-22 at 3 55 36 PM


hint: the thumbnail is the first image url in the meta images array.

Another suggestion would be to stringify the metadata before saving draft, and parse meta string on load. check `makeJsonMetadata` method from `utils/editor.ts` for that purpose.

aliseyalvi:
@noumantahir pushed update regarding the thumbnail not changing bug. 
This was a weired bug caused by a useeffect hook causing thumbIndex reset after each update in editorSettingsModal.tsx file.  
Replaced the hook with a custom function which would be called only when user changes the thumbnail. 
2nd step was to save these settings in meta which are loaded and saved when draft editor loads on ComponentDidMount in editorContainer.tsx.
3rd step was to reflect the changes on draftScreen which is referenced here in 3rd commit https://github.com/ecency/ecency-mobile/pull/2111/commits/3edd3907f0d5433cb720743f91ef86ee67d5e05e. Please check.

aliseyalvi:
https://www.loom.com/share/a742df41b8be4e8e904aed2b1fe4cb4c

noumantahir (commented):
Working good in my tests

feruzm (approved):

PR #2112, opened 2021-12-25
NT - Editor Replace

noumantahir:
### What does this PR?
After attempting to duplicate post again and again, the only possible way to duplicate is accidentally going to editor page with publish and hitting the publish button again. To avoid this scenario as well, I am replacing editor with any upcoming screen in the stack, this way user will not have a chance to duplicate post.

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/147381684-f2def210-215a-4ddb-be59-e448dc9509a8.mov
p.s. works similarly for publish as well, test with one demo post

### DIFF
--- a/src/screens/drafts/container/draftsContainer.js
+++ b/src/screens/drafts/container/draftsContainer.js
@@ -28,6 +28,8 @@ const DraftsContainer = ({ currentAccount, intl, navigation, dispatch }) => {
   const [schedules, setSchedules] = useState([]);
   const [isLoading, setIsLoading] = useState(false);
 
+  const [initialTabIndex] = useState(navigation.state?.params?.showSchedules ? 1 : 0);
+
   useEffect(() => {
     _getDrafts();
     _getSchedules();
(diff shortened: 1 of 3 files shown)


feruzm:
> ### What does this PR?
> After attempting to duplicate post again and again, the only possible way to duplicate is accidentally going to editor page with publish and hitting the publish button again. To avoid this scenario as well, I am replacing editor with any upcoming screen in the stack, this way user will not have a chance to duplicate post.
> 
> ### Screenshots/Video
> https://user-images.githubusercontent.com/6298342/147381684-f2def210-215a-4ddb-be59-e448dc9509a8.mov p.s. works similarly for publish as well, test with one demo post

We should consider moving update post button in float just like drafts, so people know when they publish and update post, distinctions are more visible. What do you think @noumantahir ?

noumantahir:
@feruzm I would suggest sticking with current implementation, it's already very clear... I did go through your discussion with user, I have enough reasons to believe that the guy might is covering his mistake.... but with this PR we have reduced those chances as well..

noumantahir:
@feruzm this needs merging as well

feruzm (approved):

PR #2113, opened 2021-12-26
updated wallet tab label to have 2 decimal figures

noumantahir:
The only difference in estimated wallet values I could find is in the wallet tab label as it was rounding the value to have to decimal point, now it has 2 for better close estimation. 

Please let me know if this was the wrong estimated value you were pointing to in trello card.

### Screenshots/Video
Screenshot 2021-12-26 at 2 22 56 PM


feruzm (approved):
This was slightly different problem/confusion but good to have this extension.

PR #2118, opened 2022-01-01
updated points estimation factor

noumantahir:
Screenshot 2022-01-01 at 9 29 49 PM
Screenshot 2022-01-01 at 9 29 43 PM

### DIFF
--- a/src/containers/pointsContainer.js
+++ b/src/containers/pointsContainer.js
@@ -211,7 +211,7 @@ const PointsContainer = ({
   const _getESTMPrice = (points) => {
     const { base, quote } = globalProps;
 
-    return points * 0.01 * (base / quote);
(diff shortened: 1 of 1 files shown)


src/containers/pointsContainer.js
```diff
@@ -211,7 +211,7 @@ const PointsContainer = ({
   const _getESTMPrice = (points) => {
     const { base, quote } = globalProps;
 
-    return points * 0.01 * (base / quote);
+    return points * 0.00442 * (base / quote);
```
feruzm, line 214: What we did in website, `points/150` that would give estimation non-related to HIVE price (base/quote) only for Boosts though. Not for Wallet points estimated value.

feruzm (approved):

PR #2119, opened 2022-01-03
Nt/entity support

noumantahir:
### What does this PR?
Added changes required in app side to support updated render-helper..

_Update_:
Now using render-helper 2.2.9

Also temporarily pointing to nt/entities-parsing branch from render-helper as without that change post body with entites is going to crash in mobile app. as soon that branch changes are merged [here](https://github.com/ecency/render-helper/pull/26), we can change back to master of render-helper.


noumantahir:
@feruzm this can now be merged perhaps

feruzm (approved):
lgtm

PR #2139, opened 2022-01-12
Nt/community link

noumantahir:
### What does this PR?
Apparently there were two bugs causing the issue, one being markdown-community-link not supported by app and other anchored image was only entertaining markdown-external-link which now reply on <a/> tag as parent to disable core image press and let anchor handle click event only.

### Issue
https://trello.com/c/kanS8VQO/309-image-link-is-not-working-as-expected

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/149085102-4102d3e4-4ec2-4c1c-834f-8a573856523d.mov


feruzm (approved):
Lgtm

PR #2143, opened 2022-01-15
Nt/local cache

noumantahir:
### What does this PR?
Introduce new store to handle vote cache, later we can use this store to handle more cached data
Purge method to discard all expired cache on app startup
Improved vote caching strategy, it is now more active than passive

The expiration time is set to 30 seconds, as more expiry time like 24 hours can lead to more confusions like what if user uses some other device to alter vote, since there is no time stamp in vote data, app will still be using local cache even though that would have been invalidated. Let me know if you think we still use 24 hours, in my test however 30 seconds works just alright for votes.

### Issue
https://trello.com/c/toQgb0Im/296-voted-caching-has-issues

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/149631853-53f1fda7-a106-4c0c-a239-1a0030cf0cc2.mov
p.s. in video there is a bug with votes count, that is also fixed in last commit I pushed to this branch


feruzm:
@noumantahir in video bug you mentioned is close at the end? I noticed that post had more votes and higher payout value when post opened but feed list had lesser, they should be consistent.

noumantahir:
@feruzm its very tricky to match perfectly for 2 reasons,

1. post registered more votes then there were when feed was loaded, the post details will show new values
2. Users own vote got registered, that is the reason I kept cache at 30 seconds as otherwise cache itself will start affecting the values.

We can spend more time mastering the cache and transaction updates, for instance we can ignore cache if user's vote is already present in post content, but even that would resolved just one problem, a vote update will still face issues. 

What do you suggest, make if perfect now or do it in iterations as we discover more.

feruzm:
@noumantahir I was referring to payout value, if I remember, we had cache update feature for feed items that would update their informational row, is it not? If it updates only on pull down to refresh, we should add logic to update that feed item values when/if post is opened.

feruzm (approved):
let's merge this for now and we can work on checking caching structure further

PR #2145, opened 2022-01-16
optimised table img width based on number of columns

noumantahir:
### What does this PR?
Recursively search for col parent if image is not a part of any columns it return maximum available width otherwise optimises width based on number of columns in a table.

Trade off is, for normal image it would still do search though it would mostly end at 1 or max 2 recursive calls for regular images which should not pose much performance hit.

I am open to suggestion as to how we can further improve the methodology.

### Issue
https://trello.com/c/EMsolR81/316-bug-table-and-images-alignment

### Screenshots/Video
Screenshot 2022-01-16 at 5 47 09 PM

Screenshot 2022-01-16 at 5 47 26 PM

Screenshot 2022-01-16 at 5 58 22 PM


feruzm (approved):
I think it is fair compromise for now, there is not much cons in extra checking, expect perhaps delay in rendering.

PR #2146, opened 2022-01-18
Nt/wallet rendering

noumantahir:
### What does this PR?
Optimised wallet page by, 
Replacing costly component with simple one in TransactionView
Replacing nested FlatList under scrollview with single FlatList
Moving extra activities filtering to side effect instead of render flow
Replaced custom Animated wallet header with CollapsibleCard

Please test is thoroughly

This is just first iteration of improving wallet page performance, I tried remove the major performance dips. But wallet have many more inefficiencies specific to api calls that we can be improved on, but left those for later as we will be redoing the wallet to newer design any ways where I rather plan on using redux more than passing props.

### Issue
https://trello.com/c/9OiKc17W/308-wallet-tab-crash-while-exploring

### Screenshots/Video

https://user-images.githubusercontent.com/6298342/149967145-8e3014fc-da3e-45b1-81da-ebce7e838d67.mov

### DIFF
--- a/src/components/basicUIElements/view/walletLineItem/walletLineItemView.js
+++ b/src/components/basicUIElements/view/walletLineItem/walletLineItemView.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import { View, Text } from 'react-native';
+import { TouchableWithoutFeedback } from 'react-native-gesture-handler';
 
 // Components
 import { DropdownButton, PopoverWrapper, Icon, GrayWrapper } from '../../..';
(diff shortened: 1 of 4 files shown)


noumantahir:
Refresh control is now functional as before.. it does fetch new data, but UI seems to not able to hold the refresh indicator, did some digging but logic seems very weird inside, can improve more tomorrow if needed.

https://user-images.githubusercontent.com/6298342/149995084-67acedb9-0c06-4f29-b4c5-1c1820566dbf.mov

feruzm (approved):
lgtm, additional improvements we can do on separate PRs, let's do some tests.

Issue #2157, opened 2022-01-23
Bug: text-justify not working


feruzm:
With new update on native rendering, `<div class="text-justify">text</div>` is not properly supported on content render.

post example: https://ecency.com/hive-142159/@dayatsiaulia/raja-si-uroe-monomad
Another similar but slightly different bug report: 
https://ecency.com/hive-125125/@liveofdalla/re-ecency-2022122t184647364z
https://discord.com/channels/385034494555455488/385448634197737484/935417936284385400

noumantahir:
After struggling with it for a while it seems the issue is coming from react-native-render-html package, I have created a comment on similar issue on their github repo. Let's hope they come back with a solution, meanwhile removing `textAlign:'justify'` prop for Android side as instead justifying it is rather clipping the like ends.

https://github.com/meliorence/react-native-render-html/issues/202#issuecomment-1022509101

noumantahir:
@feruzm 
Digging more deep revealed, it is essentially being caused because we made text selectable...it's not an issue with react-native-render-helper but coming from react-native itself...

if we remove the selectable prop, the text-justify starts working...

reference issue.
https://github.com/facebook/react-native/issues/30332

feruzm:
@noumantahir we really need text selection, paragraph selection was a compromise but even that causing issue seem bit hassle, perhaps we should rewrite some part of native-render-helper to add support for that?

noumantahir:
@feruzm how about we put it as a feature request and move forward based on what render html have to say about that...

It successful, it would fix the selectable issue on iOS, but that would still not solve the text-justify issue on android as it is rather coming from react-native itself.

noumantahir:
apprently for the current issue in hand, no address it even though it has been raised multiple times on react-native repo...

noumantahir:
closed by mistake, reopened again

feruzm:
@noumantahir Couldn't we shrink post body a bit to get around it for now?

PR #2159, opened 2022-01-27
sticking text-justify to auto for android

noumantahir:
### What does this PR?
justify textAlign with selectable enabled was doing more harm then good, the ends text was getting clipped with this set of props.

Since selectable is more important for us, for now I am ignoring justify for android so the presentation of content is not affected.

### Issue number
#2157 

### Screenshots/Video
Before
Screenshot 2022-01-27 at 1 29 09 PM

Now
Screenshot 2022-01-27 at 1 32 09 PM


feruzm (approved):
lgtm, some reported justify already was being ignored on some devices

PR #2163, opened 2022-01-28
exclusively setting div width

noumantahir:
apparently if div is present as post content parent, it was overriding the base contentWidth and taking whole available space, now setting div width exclusively to avoid this behaviour

### Steps to reproduce
https://ecency.com/hive-189157/@nerdvana/fpbvukbl

### Issue
https://trello.com/c/X8NUpc9c/321-check-post-rendering

### Screenshots/Video
Before
Screenshot 2022-01-28 at 6 44 47 PM

Now
Screenshot 2022-01-28 at 6 41 07 PM


feruzm:
@noumantahir there is conflict

feruzm (approved):
lgtm

PR #2164, opened 2022-01-28
[WIP] Getting Started Tooltip

aliseyalvi:
### What does this PR?

- Updated react-native-modal-popover module
- Created Tooltip component using react-native-modal-popover
- Added Tooltip in Editor Screen on Draft button
Work is in progress. Need further enhancements and modifications

### Issue number
Trello: Improve: Add new install draft button tooltip in editor screen

### Screenshots/Video
https://www.loom.com/share/49e55b781f924b4fa546a691e853c6ce


aliseyalvi:
@noumantahir registered tooltip in redux when it loads for first time. Checkout the latest update  https://github.com/ecency/ecency-mobile/pull/2164/commits/3f2b0d968d0dcf23e1468799098ac2836e8c1b51
https://www.loom.com/share/d32a4965dba641b0b377f7506e22c558

src/components/markdownEditor/view/markdownEditorView.js
```diff
@@ -299,17 +301,25 @@ const MarkdownEditorView = ({
...
-            isLoading={isLoading}
-          />
-        </AnimatedView>
+        <>
+          <AnimatedView
+            style={styles.floatingContainer}
+            animation="bounceInRight"
+            onAnimationEnd={() => tooltipRef.current?.openTooltip()}
+          >
+            <Tooltip ref={tooltipRef} text="This is Draft" walkthroughId="DRAFT_BTN_TOOLTIP">
```
feruzm, line 310: String should be in Translation files

noumantahir (changes requested):
The requested changes are tweaked towards making the structure scaleable as we move forward. Though tight now we only have tooltip for drafts but we should still structure it so we face minimal roadblocks as we scale

aliseyalvi:
@noumantahir updated redux store according to your review. Storing walkthrough indices inside map and keys in constants file. Registering a tooltip is a bit different process. Actually controlling this process from ref object so that parent has access to the method. 
Latest commit https://github.com/ecency/ecency-mobile/pull/2164/commits/f408db9979de0d53102401d998c4fdfe0479a1d9

src/redux/reducers/index.js
```diff
@@ -21,4 +22,5 @@ export default combineReducers({
   communities,
   user,
   cache: cacheReducer,
+  tooltips: tooltipsReducer,
```
noumantahir, line 25: @aliseyalvi  perhaps renaming `tooltipsReducer` / actions to `walkthroughReducer`/action will make more self elaborative, what do you think?

src/redux/reducers/tooltipsReducer.ts
```diff
@@ -0,0 +1,30 @@
+import { REGISTER_TOOLTIP } from '../constants/constants';
+
+export interface Walkthrough {
```
noumantahir, line 3: suggestion: perhaps should be renamed to `WalkthroughItem` ?

src/config/locales/en-US.json
```diff
@@ -340,7 +340,8 @@
     "done":"DONE",
     "draft_save_title":"Saving Draft",
     "draft_update":"Update current draft",
-    "draft_save_new":"Save as new draft"
+    "draft_save_new":"Save as new draft",
+    "load_draft_tooltip": "Load your lastest draft here"
```
noumantahir, line 344: Let's introduce a separate section where all walkthrough related texts reside so it's easier to manage them in one place later...? moving forward we can even devise some mechanism to pair them with walkthrough ids, but not now.

```
"walkthrough":{
   "load_draft":"asdfasldkf"
}
```

noumantahir (changes requested):
I reviewed the code changed, I am glad we are moving the right direction, just a few more changes and we should be good.

However, functionality wise I was unable to test the app as it crashed as soon I entered the editor screen, apparently we need a null check for walkthrough map but cannot say for sure.. What I can say is redux crashes in this way when an older version of reducer is already present in existing app.

Screenshot 2022-02-09 at 11 22 01 AM


Update: tried a new install, app still crashed but with a different error this time....

Screenshot 2022-02-09 at 11 34 34 AM

aliseyalvi:
@noumantahir added tarnsform function for walkthroughRedcuer and renamed reducer and actions. Used map in reducer and I think it is flexible enough to implement the walkhrough for the app.
Check latest update here https://github.com/ecency/ecency-mobile/pull/2164/commits/bdfa973e90c5171ef8d6fe6b6957a36bb7752527

noumantahir (approved):
lgtm

PR #2166, opened 2022-01-28
Nt/comments cache

noumantahir:
### What does this PR?
renders cached comment if comment is made but not yet returned by api request. right now caching is only supported if made suing quick reply modal

In later code updates will add support for comment update caches and caching support using full editor. 

**Update**
Made overall comment cache experience more stable and smart
Added support for comment editing while utilising cache

**Update**
PR can now be tested

Things need checking or pending

- [x] function of comment actions if rendered from cache
- [x] working of reply to cached comment
- [x] rendering of cached reply to a comment
- [x] opening thread of cached comment
- [x] handling comment updates using cache 
- [x] editing a cached comment
- [x] deleting a cached comment


noumantahir:
@feruzm any feedback on this... it's pending for quite some time now.

feruzm (approved):
lgtm, will need to test this more.

PR #2167, opened 2022-01-29
fixes misfired user search 

noumantahir:
Instead of firing debounce on trailing edge by default, we now run on leading edge, this not just makes sure avoid delayed misfire but also improves the regular user experience. Also added debounce cancellation for added redundancy. 

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/151655872-98ae2023-8784-4fff-8af7-b0f96e28bde5.mov


feruzm (approved):
Lgtm

PR #2171, opened 2022-01-30
updated share content props for refering

noumantahir:
### What does this PR?
Apparently android do not recognise shared url, yet message prop is the one both platforms recognise

### Screenshots/Video
Screenshot 2022-01-30 at 9 50 47 PM


feruzm (approved):
lgtm

PR #2187, opened 2022-02-09
React Native Video for Extension based video URLs

### LINKED ISSUE #2177: [IMPROVE] using react-native-video for extension based video streaming links
Right now we use webview for streaming extension based video urls that in some way affect the user experience as such urls are wrapped to out custom code to avoid autoplay, but the drawback is that we lose the thumbnail of video.

https://github.com/ecency/ecency-mobile/blob/48518bb90c9de13e3c929ff83f1662dd1bc99cd1/src/components/postHtmlRenderer/postHtmlRenderer.tsx#L226

Perhaps a better solution would be to use react-native-video package to stream such urls. that would also add benefit to not limited to using webview for video playbacks.

**To Reproduce**
1. Go to this ['Post'](https://ecency.com/hive-172868/@vaketo/-----xgo102be5k8xjugq0drdwpsg4rjvu0gk)
2. Check video playback experience.
3. The thumbnails are blank before starting playback

**Expected behaviour**
We should try to render thumbnails as much as possible before user hits the play button on videos.

**Screenshots or Video**
Screenshot 2022-02-03 at 3 54 30 PM

**Additional context**
What I suggest we do is, we pass the uri directly to VideoPlayer in url mode, the VideoPlayer process the url for extension and render either native video player or webview based player depending on extension returned.

https://github.com/ecency/ecency-mobile/blob/48518bb90c9de13e3c929ff83f1662dd1bc99cd1/src/components/videoPlayer/videoPlayerView.tsx#L9

aliseyalvi:
### What does this PR?
fixes #2177 
Added react-native-video module for playing extension based video urls. 
Solves the issue of thumbmail not showing.

### Issue number
#2177 

### Screenshots/Video
https://loom.com/share/4067f50c0b604674b0e24fc64dc6a2d9


src/components/videoPlayer/videoPlayerView.tsx
```diff
@@ -1,37 +1,35 @@
...
 interface VideoPlayerProps {
-  mode: 'source'|'youtube'|'url';
+  mode: 'source' | 'youtube';
   contentWidth?: number;
   youtubeVideoId?: string;
-  videoUrl?: string;
   startTime?: number;
-  source?: WebViewSource;
-
+  source?: string;
```
noumantahir, line 13: let's rename `source` to `uri`, as source prop name can be misleading.

noumantahir (changes requested):
Exactly how I expected it to behave....just one tiny change requested.

src/components/videoPlayer/videoPlayerView.tsx
noumantahir: and same goes for available values of `mode` prop

aliseyalvi: @noumantahir updated the props.

noumantahir:
@aliseyalvi iOS side is working great. do also test playback on android side... in my tests it is not behaving as expected...

note: on android, we are sticking with modal based player for now, react-native-video should atleast play video in modal. that being said, I also force tested it in android inline playback, seem it is greatly misbehaving, for instance, not video thumb show up and playback controls float on screen.
Screenshot 2022-02-14 at 3 05 54 PM

aliseyalvi:
@noumantahir please check the latest update here https://github.com/ecency/ecency-mobile/pull/2187/commits/176fa12a01cdb0c90694116f8361ac33582c8bb8.

- Added react-native-media-controls for showing controls in android because on android the controls are doesn't supported and default controls greatly misbehave. Using custom controls also add further flaxibility for adding additional controls later.
- on iOS side working good in inline and on android tested inside modal. A minor bug on android side is that player deosn't support full screen in android. This is a bug in react native video. Force tested as well. To fix this we would have to make this behaviour custom i-e extend the modal heights and widths to cover the whole screen and showing the close button on top to dismiss full screen.
-

noumantahir (approved):

noumantahir:
@aliseyalvi working as expected, tested on android as well....

Screenshot 2022-02-15 at 1 03 39 PM
Screenshot 2022-02-15 at 1 04 44 PM

@feruzm you may review and merge now....

feruzm (approved):
Lgtm

PR #2194, opened 2022-02-18
Nt/lower case url

noumantahir:
### What does this PR?
this pr handle two places where url may have capital permlink, let me know if you think we can reduce this to one...

1. before fetching post
2. while parsing deeplink url

### Screenshots/Video
Screenshot 2022-02-18 at 10 25 02 PM


feruzm (approved):
lgtm

PR #2195, opened 2022-02-19
add marking hive notifications

feruzm:
This should mark Hive notifications as read on top of our internal notifications marking. Non tested, just integrated, please test.
https://github.com/orgs/ecency/projects/2#card-77970890

### DIFF
--- a/src/providers/hive/dhive.js
+++ b/src/providers/hive/dhive.js
@@ -1236,8 +1236,8 @@ export const unfollowUser = async (currentAccount, pin, data) => {
   );
 };
 
-export const markHiveNotifications = async (currentAccount, pin) => {
-  const digitPinCode = getDigitPinCode(pin);
+export const markHiveNotifications = async (currentAccount, pinHash) => {
+  const digitPinCode = getDigitPinCode(pinHash);
   const key = getAnyPrivateKey(currentAccount.local, digitPinCode);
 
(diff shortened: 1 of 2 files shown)


noumantahir (commented):
@feruzm the method fails while trying to process pin code....I can look into it's details

Screenshot 2022-02-20 at 10 00 15 PM
Screenshot 2022-02-20 at 10 00 24 PM

feruzm:
@noumantahir please do, I just assumed those props are passed on or available, not sure for certain if it is available in props.

noumantahir:
Update:

@feruzm tested call for succession, though I am not sure how I can further test this or what exactly it does to more specific in testing it's working.

Screenshot 2022-02-20 at 10 09 42 PM

Screenshot 2022-02-20 at 10 10 32 PM

Also regarding it's resemblance with `markNotification`, I checked the implementations or `markNotifications`, it is also being used in `applicationContainer`, do you think `markHiveNotification` also needs to be implemented there or just required in one place that you implemented.

Screenshot 2022-02-20 at 10 14 34 PM

feruzm:
@noumantahir applicationContainer one likely when opening notifications from push and it marks particular item, not all, so we don't need to integrate this there.

noumantahir (approved):
all good then I guess....

PR #2197, opened 2022-02-22
Sa/refer screen

aliseyalvi:
### What does this PR?
- Fixes the popover arrow issue in refer screen
- Open quick profile modal on user tap


feruzm (approved):
lgtm

PR #2200, opened 2022-02-25
[WIP] QR Screen 

### LINKED ISSUE #2181: Feature: QR scan
On side menu have item for QR code, when clicked open screen with:

1. Ability to scan QR
2. Show QR with encoding username of logged in account if exist. This can be used for receiving funds.
3. On Transfer screens, `To` field/input should have QR icon next to it for scanning and quickly filling username.
4. Add logic for parsing QR encoding of different logic.
For example: `username:transfer` would open transfer screen with `To` field using `username`. `username:view` would open Profile page of `username`.  `username:boost` would open purchasing account boosting feature which we already have implemented.

aliseyalvi:
### What does this PR?

- Added QR screen
- Added Basic QR Scanner
- Added url handler. Currently it handles only valid url. Will have to add validation checks for url
- No option for rescanning. Will have to add it as well

### Issue number
fixes #2181


src/screens/qrScreen/qrScreen.tsx
```diff
@@ -0,0 +1,177 @@
...
+    setScanResult(e.data);
+    setIsActive(false);
+  };
+
+  const _handleOpen = () => {
+    console.log('scanResult : ', scanResult);
+    _handleDeepLink(scanResult)
+  };
+
+  const _handleDeepLink = async (url = '') => {
```
noumantahir, line 35: @aliseyalvi  We should reuse current implementation... perhaps more it to utils and use same method from both components.

src/screens/qrScreen/qrScreen.tsx
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    if (routeName && keey) {
+      navigate({
+        routeName,
+        params,
+        key: keey,
+      });
+    }else{
+      console.log('unable to open url');
```
noumantahir, line 123: This section can be used to return validity... we can repurpose this method to return routeName, params and key... if these properties exist in returning object, then show validity tick mark otherwise show cross.

Used the same returned data on open button press.

@aliseyalvi this should make more sense, feel free to reach out for any queries
```
const [navData, setNavData] = useState(null);
const [readUrl, setReadUrl] = useState('');
.
.
.

const navData = _handleDeepLink(url);
setIsValid(navData?true:false);
.
.
.
const _openPressed = () => {
    navigate(navData);
}
.
.
.
<>
{
   readUrl && <ValidTick isValid={navData}/>
}
</>
```

noumantahir (commented):
@aliseyalvi left 2 review comments, do check them out

aliseyalvi:
- added deepLinkParser util method.
- validated scanned url and show tick/cross based on validation.
https://github.com/ecency/ecency-mobile/pull/2200/commits/6e84ee5f37fd4750cab07f4b5e4002c1a8824a83

aliseyalvi:
fixed camera overflow issue caused by increasing camera width https://github.com/ecency/ecency-mobile/pull/2200/commits/1fd15d305f20ad8b0801aebe204651459b3a37da

aliseyalvi:
@noumantahir 
-added full screen QR modal
-auto navigate to url if recognizes successfully
-show alert if unable to recognize url

src/redux/store/store.ts
```diff
@@ -35,11 +36,12 @@ const persistConfig = {
   // Blacklist (Don't Save Specific Reducers)
   blacklist: ['nav', 'application', 'communities', 'user'],
   timeout: 0,
-  transforms:[transformCacheVoteMap,transformWalkthroughMap]
+  transforms:[transformCacheVoteMap,transformWalkthroughMap],
+  stateReconciler: autoMergeLevel2,
```
noumantahir, line 40: What does this do....?

src/components/qrModal/qrModalView.tsx
```diff
@@ -0,0 +1,106 @@
...
+    setIsProcessing(true);
+    const deepLinkData = await deepLinkParser(url, currentAccount);
+    const { routeName, params, key } = deepLinkData || {};
+    setIsProcessing(false);
+    if (routeName && params && key) {
+      setIsScannerActive(false);
+      _onClose();
+      navigate(deepLinkData);
+    } else {
+      Alert.alert('Unsupported URL!', 'Please scan a valid ecency url.', [
```
noumantahir, line 57: Hmm... Alert may be too intrusive.... Also, message should contain QR in place of url, like Unsupported QR code

noumantahir (commented):
Left a few comments, reviewed from phone could not try things out

src/redux/store/store.ts
aliseyalvi: actually I was struggling with issue which I have mentioned with you earlier as well. The redux store wasn't reflecting changes when I try to change the reducers structure or add some property/dispatch method. Redux persist has default stateReconciler to autoMergeLevel1 , I had changed it to autoMergeLevel2 so it merges the new changed state and already saved state. U can refer to redux-persist docs here https://github.com/rt2zz/redux-persist.

src/components/qrModal/qrModalView.tsx
aliseyalvi: I added alert because I checked whatsapp and they are also showing alert message for invalid QRs, u can also suggest some nice way to show the message.

src/redux/store/store.ts
noumantahir: I am entirely sure how it might affect the overall app behaviour internally... @feruzm any thoughts on this.

@aliseyalvi as for reducers to change in persist we usually use spread operator to achieve that, {...state}

src/components/qrModal/qrModalView.tsx
noumantahir: Okay, we can keep it for now...

src/redux/store/store.ts
feruzm: @noumantahir , @aliseyalvi I checked docs, it says `This acts just like autoMergeLevel1, except it shallow merges two levels`. I don’t think it would break anything but we need to check cache and other recent additions and make sure we won’t have infinite render loops somewhere. Sadaqat maybe go through each page or redux usages quickly and see if you notice something that breaks.

aliseyalvi: Ok, i'll check on my side  as well when checkout the branch, better if u can check on your side by checking this branch, and let me know if anything weired happens.

feruzm (approved):
Looks good for initial phase 👍

PR #2210, opened 2022-03-01
[WIP] RN Upgrade 0.63 -> 0.70.3

noumantahir:
### What does this PR?
upgraded RN to 0.67 to a point apps are building successfully. At this point we are having static startup screen on both iOS and Android as shown in screen shot below. There is no error, there is not crash or report etc.

Only we have warning related to `NativeEventEmitter() `but we are not calling them from app ourselves

In case you intend to try it out, after `pod install` make sure to remove `GCDAsyncSocket.m` from `Pods -> Build Phases -> Compile Sources`. Will come up with an automated solution before finalising

Screenshot 2022-03-01 at 6 04 22 PM 1

### Steps to reproduce
- `yarn`
- `cd ios && pod install `
- remove specified file from compile sources as mentioned above.
- `npx react-native run-ios`

### Note To Self

- [x] Test coin details actions not working as expected
- [x] media uploads modal swipe gesture not responding

### Items postponed for end

- [x] Build and test iOS
- [x] Implement react-native-bootsplash package
- [x] integrate notifee package
- [x] upgrade react-native-reanimated


noumantahir:
@feruzm at this point we have stabilised our app with 0.67, instead of starting from scratch, I resume this penning upgrade PR to save time and it was worth it.

The app itself is roughly 90% stable, there some feature that are not working but the ones left are mostly minor and easy to fix.

Now we have two options, either we can upgrade to 0.68 that cover one of editText crash fix from react-native side, or we can go all the way to 0.7 that adds another EditText crash issue fix but also utilises new rendering architecture https://reactnative.dev/architecture/fabric-renderer

upgrade to 0.68 should not be an issue as there is hardly a month long gap between our current version 0.67.3 and 0.68.  but 0.7 can be a real deal.

Let me know what you suggest.

noumantahir:
> 

relevant editText PRs share similar title but the code itself is quite different also the gap between PR creations dates is also huge, so I am assuming they cover different aspects

feruzm:
@noumantahir Ok, How complex upgrade to 0.7 at this point? It automatically includes fixes in 0.6x, I assume.

noumantahir:
@feruzm  We should be good merging this PR with development branch...issues we are resolving now in this branch are not directly related to rn upgrade

feruzm (approved):
great work and elegant solution with patches 👏 🙇

PR #2212, opened 2022-03-02
Fix Notifications scroll bug

### LINKED ISSUE #2206: Notification auto scroll
**Report**:
When we open a comment, the list of comments scroll position is change upon coming back from comment details

**Assessment**:
Apparently, when user open the comment, app tries to mark that comment as read, upon successful api operation, the notification are fetched again probably as a result of props update. It could be something else but need more deep checking

https://github.com/ecency/ecency-mobile/blob/3f64c46268b3e6f94da2c4def02d7e893aa6ae2e/src/screens/notification/container/notificationContainer.js#L89

https://github.com/ecency/ecency-mobile/blob/3f64c46268b3e6f94da2c4def02d7e893aa6ae2e/src/screens/notification/container/notificationContainer.js#L179

Screenshot 2022-02-28 at 5 19 34 PM

**Ref**:
https://discord.com/channels/@me/920267778190086205/947462092154368050

aliseyalvi:
### What does this PR?
This fixes #2206 ,  the notifications scroll bug when user open a notification and come back the scroll position changes. 
### Issue number
#2206

### DIFF
--- a/src/screens/notification/container/notificationContainer.js
+++ b/src/screens/notification/container/notificationContainer.js
@@ -169,20 +169,18 @@ class NotificationContainer extends Component {
     await this.setState({ selectedFilter: value, endOfNotification: false, selectedIndex: ind });
   };
 
-  // Note: This snippet is causing double api call which cause scroll position reset
-  /*
-    UNSAFE_componentWillReceiveProps(nextProps) {
+  UNSAFE_componentWillReceiveProps(nextProps) {
     const { selectedFilter } = this.state;
     const { currentAccount } = this.props;
-
     if (
-      (nextProps.activeBottomTab === ROUTES.TABBAR.NOTIFICATION && nextProps.currentAccount.name) ||
-      (nextProps.currentAccount.name !== currentAccount.name && nextProps.currentAccount.name)
+      currentAccount &&
+      nextProps.currentAccount &&
+      nextProps.currentAccount.name !== currentAccount.name
     ) {
       this.setState({ endOfNotification: false }, () => this._getActivities(selectedFilter));
     }
   }
- */
+
   render() {
     const { isLoggedIn } = this.props;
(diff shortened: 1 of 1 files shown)


src/screens/notification/container/notificationContainer.js
```diff
@@ -168,7 +169,9 @@ class NotificationContainer extends Component {
     await this.setState({ selectedFilter: value, endOfNotification: false, selectedIndex: ind });
   };
 
-  UNSAFE_componentWillReceiveProps(nextProps) {
+  // Note: This snippet is causing double api call which cause scroll position reset
+  /*
+    UNSAFE_componentWillReceiveProps(nextProps) {
     const { selectedFilter } = this.state;
     const { currentAccount } = this.props;
```
noumantahir, line 177: @aliseyalvi the logic is checking if user account changes, refetch the comments... I think the logic is not setup write, check how there is an OR statement with logic just checking if current tab is Notifications tab and user is logged in... which means, activities will be refetched on every prop update...

The logic should simply by... if account change then refresh... no need to check tab or anything else.....because even if tab is not notification we will still have to fetch new notifications irrespective...

long story short....try
`if(currentAccount && nextProps.currentAccount && nextProps.currentAccount.name !== currentAccount.name)`

p.s. make sure to test account change and logging out as well.

noumantahir (changes requested):

src/screens/notification/container/notificationContainer.js
aliseyalvi: @noumantahir added condition for account change and test as well. working good now. check update here https://github.com/ecency/ecency-mobile/pull/2212/commits/d7034518003562f521981ed06ed8f5ec2ae5f8d7

noumantahir (approved):

noumantahir:
@feruzm it's good from my side

PR #2218, opened 2022-03-10
fixes text clipping in transfer screen

aliseyalvi:
### What does this PR?
This fixes the text clipping in transfer screens. 
image

### Steps to reproduce
Go to wallet
then go to hbd transfer screen
in the last, text is clipped

### Screenshots/Video
image


feruzm (approved):
lgtm

PR #2219, opened 2022-03-13
Fix comment delete failure bug

### LINKED ISSUE #2168: [BUG] Comment delete failure
**Describe the bug**
While trying to delete multiple comments comment_cashout error appeared on deleting about 5th comment.

**To Reproduce**
1. Go to post comments with a number of your comments
2. Click on delete for each
4. Delete fails after a few deletions

**Expected behavior**
There should be a user friendly way to representing the rejection.

**Screenshots or Video**
Screenshot 2022-01-29 at 2 45 40 PM

 - Device: iPhone 11
 - OS: 15.0
 - Version: 3.0.26

aliseyalvi:
### What does this PR?
fixes #2168. 
This PR fixes the comment delete failure bug. It only shows the delete button when delete conditions are met.

### Issue number
#2168 
### Screenshots/Video
image

### DIFF
--- a/src/components/comment/view/commentView.tsx
+++ b/src/components/comment/view/commentView.tsx
@@ -233,7 +233,7 @@ const CommentView = ({
               onPress={() => handleOnEditPress && handleOnEditPress(comment)}
               iconType="MaterialIcons"
             />
-            {!childCount && !activeVotes.length && (
+            {!childCount && !activeVotes.length && isCommentDeletable && (
               <Fragment>
                 <IconButton
                   size={20}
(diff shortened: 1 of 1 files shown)


aliseyalvi:
@noumantahir there is a glitch as well. When user edits the comment it shows the delete button that is because after edit the last edit comment returns very less data to process the condition resulting in failing to check the condition. The issue seems due to caching of comment because when we go back and come again on post it fetches comments again and fixes the glitch.

noumantahir:
@aliseyalvi definitely coming from cache, because we do not cache every comment detail... but we surely can....

feruzm (approved):
Lgtm

PR #2228, opened 2022-03-15
using touchable from react-native instead of gesture handler

noumantahir:
Fixed thumb selection issues, oddly enough, I tried changing the same in delegateScreen and the touchable rather stopped working, there is something more to it that we do not understand, in some places touchableOpacity from gesture-handler work and in others default react-native one works.

### Screenshots/Video
Screenshot 2022-03-16 at 3 13 47 AM


feruzm (approved):
let's try

Issue #2147, opened 2022-01-19
Bug: Landscape and portrait crashes app


feruzm:
I have noticed this before on mobile app. It seems if I watch a video and try full screen it in portrait it works but if I try turn the screen for a full screen video then the Ecency app freezes and I need to restart it.

aliseyalvi:
Duplicate issue. Refer here https://github.com/ecency/ecency-mobile/issues/2093#issuecomment-1065894479

Issue #2128, opened 2022-01-06
Bug: Youtube embed landscape


feruzm:
In ecency app i cannot look yt -video in landscape. If i select full scren and flip my mobile ecency restarts. Is thus known issu?

Galaxy S21 5G,  Android 11

aliseyalvi:
Issue is due to landscape support is not yet added in ecency. We would have to support landscape step by step just like we have added support for RTL languages.

aliseyalvi:
duplicate issue. Refer here https://github.com/ecency/ecency-mobile/issues/2093#issuecomment-1065894479

Issue #2093, opened 2021-11-17
[BUG]


unklebonehead:
**Describe the bug**
While trying to full screen a 3Speak video and rotate to landscapt to view app freezes with only Ecency logo showing. Locks phone up and required a soft reboot to fix.

**To Reproduce**
While watching a 3 Speak video tap the full screen icon and turn phone to landscape. (On it's side)

**Expected behavior**
Expected to be able to watch 3Speak video full screen.

**Screenshots or Video**
Unable to get screenshot due to phone frozen..

**Smartphone (please complete the following information):**
 - Device: OnePlusOne 6T
 - OS: OxygenOS (OnePlusOne Android)
 - Version 10.3.12 (Android 10)

**Additional context**

Can provide logs if you need.

aliseyalvi:
currently landscape is not supported in ecency. Landscape support needs to be added to fully support the fullscreen videos in landscape.

PR #2236, opened 2022-03-26
Landscape support

aliseyalvi:
### What does this PR?
This PR enables landscape mode in Ecency.  
It enables the landscape support in the whole app and uses react-native-orientation-locker package to lock the orientation to portrait in application screen. Orientation is only unlocked in post screen and locked again on coming back. From this point we can add support for landscape in each screen and unlock the orientation lock in that screen.

### Steps to reproduce
On feed screen enable rotation and turn screen to landscape. App would not crash as it crashes previously. But it would remain in portrait due to orientation lock. 
Now tap on any post and go to post screen. Turn screen to landscape again and you will see app in landscape mode with some distorted layout. 

### Issue number
#2093. #2147 . #2128. 
### Screenshots/Video

Screenshot_20220326-181335
Screenshot_20220326-181406

**UPDATE**
landscape content is now centred and have width equivalent of portrait width to avoid re-render.
Screenshot 2022-04-06 at 11 50 54 PM


aliseyalvi:
@noumantahir please check the latest changes.

- center aligned the post body when landscape is active.
- changed sticky tab bar height and padding

https://user-images.githubusercontent.com/48380998/161905859-a6458dd2-2166-4957-afae-486071abf63c.mp4

src/components/postView/view/postDisplayView.js
```diff
@@ -20,8 +22,10 @@ import { ParentPost } from '../../parentPost';
 // Styles
 import styles from './postDisplayStyles';
 import { OptionsModal } from '../../atoms';
+import { orientations } from '../../../redux/constants/orientationsConstants';
 
 const HEIGHT = Dimensions.get('window').width;
```
noumantahir, line 27: @aliseyalvi height is equal to width here, could be a typo

src/components/postView/view/postDisplayView.js
```diff
@@ -50,8 +56,21 @@ const PostDisplayView = ({
...
+  const [contentWidth, setContentWidth] = useState(WIDTH - 32);
+  const deviceOrientation = useSelector((state) => state.ui.deviceOrientation);
 
   // Component Life Cycles
+  useEffect(() => {
+    if (
+      deviceOrientation === orientations.LANDSCAPE_LEFT ||
+      deviceOrientation === orientations.LANDSCAPE_RIGHT
+    ) {
+      setContentWidth(Dimensions.get('window').height - 128);
```
noumantahir, line 68: @aliseyalvi shouldn't this change since now we do not want to change contentWidth

noumantahir (changes requested):
Added code comments, and this UI glitch needs testing.

Overall, this seems to have fixed the packager reset problem, tested on a 2014 phone emulator, do not have any that much old real device to test

Screenshot 2022-04-06 at 11 06 51 PM

src/components/postView/view/postDisplayView.js
aliseyalvi: yes, will change that

src/components/postView/view/postDisplayView.js
aliseyalvi: it coule de a typo

noumantahir (approved):

noumantahir:
Everything seems on point from my side, @good-karma looking forward to your feedback....

feruzm (approved):
Lgtm, need some more testing on different devices

PR #2239, opened 2022-03-27
Nt/wallet fixes

noumantahir:
### What does this PR?
fixed issue causing profile wallet data to fail, we can improve that section of wallet as well moving forward. for now, using legacy wallet.

### Screenshots/Video
Screenshot 2022-03-28 at 12 00 39 AM


feruzm (approved):
lgtm

PR #2255, opened 2022-04-13
Nt / No user crash

noumantahir:
### What does this PR?
App was crashing if there was no logged in user after adding support for blog pinned post


feruzm:
Added into recent pinning PR? @noumantahir

PR #2260, opened 2022-04-14
Nt/smooth notification content

### LINKED ISSUE #2149: Improve: notification tabs content loading
Right now each tab content is fetched every time tab is changed, the idea is just like we maintain state of feed posts we can do the same for notifications as well.

And make experience even better, we cache some last fetched notifications and populate them first as user enter notifications tab, just like default feed tab.

Moreover, we can periodically, or based on notification event, look for new notifications and show a popup like we do for new posts.. `X New items` etc

https://user-images.githubusercontent.com/6298342/150194957-89174d47-d936-4807-b500-55e8d7f36fd8.mov

noumantahir:
### What does this PR?
This is first iteration of notification tabs content improvement, but already it's substantial improvement. 

### Note
The experience can be further improved by maintaining separate tabs state and integrated cache to fetch first notification page on startup and show that. The overall experience can be made as smooth as primary posts tab experience. 

But before jumping into that need to assess if it's worth spending that much time improving this experience, after experience the current update, only this PR iteration should seems good enough.

@feruzm need your thoughts.

Explore notification filters

### Issue number
fixes #2149 

### Screenshots/Video
**Before**
https://user-images.githubusercontent.com/6298342/163291378-c19c4367-ebfc-4326-b10e-4a2070fea017.mov

**After**
https://user-images.githubusercontent.com/6298342/163291407-4dc65dd6-c5fe-4c77-a759-c583b0aa6abc.mov

### DIFF
--- a/src/components/notification/view/notificationView.tsx
+++ b/src/components/notification/view/notificationView.tsx
@@ -24,6 +24,8 @@ class NotificationView extends PureComponent {
    * ------------------------------------------------
    *   @prop { type }    name                - Description....
    */
+  listRef = null;
+
   constructor(props) {
     super(props);
     this.state = {
@@ -37,6 +39,7 @@ class NotificationView extends PureComponent {
       selectedFilter: 'activities',
       selectedIndex: 0,
     };
+    this.listRef = React.createRef();
   }
 
   // Component Life Cycles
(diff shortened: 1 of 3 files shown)


feruzm (approved):
looks good to me, we replaced animations with lottie before perhaps we should try that here as well for better performance?

PR #2264, opened 2022-04-15
Nt/auto detect theme

noumantahir:
### What does this PR?
using appearance module to detect current theme.
added native side fix to detect theme change on the fly on android.
tested on both android and ios.

current behaviour is now similar to what we had before.


feruzm (approved):
Lgtm

PR #2266, opened 2022-04-15
Nt/auto detect theme

### LINKED ISSUE #2175: [IMPROVE] put back automatic color scheme detected
**Describe the bug**
A few months back we removed theme detection because of some error, it seems we can now user useColorScheme from react-native put back the support.

https://reactnative.dev/docs/usecolorscheme

**Additional context**
Currently we use redux state flag to determine theme, we can keep using the structure, just update the flag based on useColorScheme value but prioritise user selected them from setting.

FYI: the hook always returns light if debugger is enabled, it's a known bug that should not affect production

noumantahir:
### What does this PR?
- we now support light dark and system defined theme. 
- while running in background, app theme will only change if it is set to system defined.

### Issue number
fixes #2175 

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/163634937-776daa7e-ce7c-4adf-af6c-5bc43143fc1f.mov

### DIFF
--- a/android/app/src/main/java/app/esteem/mobile/android/MainActivity.java
+++ b/android/app/src/main/java/app/esteem/mobile/android/MainActivity.java
@@ -33,6 +33,7 @@ public class MainActivity extends ReactActivity {
     setIntent(intent);
   }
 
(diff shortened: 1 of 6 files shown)


feruzm (approved):
lgtm

PR #2279, opened 2022-04-27
Scroll Glitch

noumantahir:
### What does this PR?
handling hyperlink image press using custom image component

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/165411696-3e822575-a756-46ea-aba5-578b1623eeae.mov

Screenshot 2022-04-27 at 4 57 55 AM

### Blocker
This PR attempts handle glitch only when hyperlink have one child and that is an image. However glitch is still there when rendering other hyperlinks as they are still handled by Render Html component as shown in demo video as well.

The issue is, we do not know how complex can a hyper link get, if it contains custom text and content with either deep or multiple children, it will become very difficult to process them.

That being said, the issue is very minimal in my opinion and is something that would occur rarely, the image one however was quite considerable and can be slightly improved as well.

@feruzm what are your thoughts on this.

**Blocker Update**
We should be able to handle this type of no children hyperlink as well, possible issue is we can lose parent font properties passing on to hyperlink.

Screenshot 2022-04-27 at 5 08 46 AM

If we use onPress method of <Text the glitch does not go away, but if we wrap it in some touchable, the glitch go away however we lose text integrity, check screenshot below.... spend about 30 minutes but could not figure it out so far.

Screenshot 2022-04-27 at 5 46 03 AM

### DIFF
--- a/src/components/autoHeightImage/autoHeightImage.tsx
+++ b/src/components/autoHeightImage/autoHeightImage.tsx
@@ -50,7 +50,7 @@ import { TouchableOpacity } from "react-native-gesture-handler";
     }
 
     return (
-      <TouchableOpacity onPress={onPress} disabled={isAnchored} activeOpacity={activeOpacity}>
+      <TouchableOpacity onPress={onPress} disabled={isAnchored} activeOpacity={activeOpacity || 1}>
         <FastImage 
           style={imgStyle}
           source={{uri:imgUrl}}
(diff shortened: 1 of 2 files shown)


feruzm:
@noumantahir I am not sure, mobile app having some limits in functionality is ok I think but I don't to complicate the logic if it is hard to maintain and improve in future. I am ok with leaving issue as is for now.

noumantahir:
@feruzm  **Update:**
- Added slight activeOpacity for anchored image to differentiate between regular and anchored image
- Dropped deep processing raw hyperlinks for now to avoid glitch for now

https://user-images.githubusercontent.com/6298342/165868657-ed99357a-4e75-445a-a3a4-0e02db5e58a9.mov

feruzm (approved):
Lgtm

PR #2303, opened 2022-05-13
skip access token injection if already present

noumantahir:
### What does this PR??
If user had two logged in users, this code was mixing access tokens resulting in a wrong response on other user login.


feruzm (approved):
Lgtm

PR #2307, opened 2022-05-15
Nt/points api updates

noumantahir:
### What does this PR?
- updated ePoint api's getUser method, and renamed to getPointsSummary for better meaning
- updated method name from getUserPoints to getPointsHistory
- wrapped optional current account values in try catch, so if one failed app can still login
- it should also potentially fix the 404 error, possibly.


feruzm (approved):
Lgtm, let’s do some tests

PR #2308, opened 2022-05-15
Sa/orientation locker bug

aliseyalvi:
### What does this PR?

- This PR disbales the landscape lock and unlock and locks screen always to portrait. This would disable landscape but it will atleast do not crashes when landscape is activated in device or app is forced to landscape.
- Second thing it unlocks orientation when video player is set to full screen but this functionality is limited espacially on android. That is because we are using multiple methods to play videos like we are using react-native-video, react-native-webview and react-native-youtube-iframe to play different type of videos. Only react-native-video supports full screen change event and react-native-youtube-iframe partially detects it on android Here is [Issue](https://lonelycpp.github.io/react-native-youtube-iframe/component-props/#onfullscreenchange). In webview we dont have method to detect if it is playing in fullscreen .

- Also it needs more testing for different formats and type of videos and different platforms needs more testing as well.

Go to any post with some video.
Play the video
try to full screen the vide and change device orientation and check if orientation changes or it is fixed to portrait. 
It will show mix behaviour. Static video files should support landscape on both android and iOS. Youtube videos should also support landscape on android. 

### Issue number
https://discord.com/channels/@me/920267778190086205/975297503928926209

### DIFF
--- a/src/components/videoPlayer/videoPlayerView.tsx
+++ b/src/components/videoPlayer/videoPlayerView.tsx
@@ -1,11 +1,12 @@
-import React, { useState, useRef } from 'react';
+import React, { useState, useRef, useEffect } from 'react';
 import { Dimensions } from 'react-native';
 import { View, StyleSheet, ActivityIndicator } from 'react-native';
 import WebView from 'react-native-webview';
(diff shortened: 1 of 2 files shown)


feruzm (approved):
lgtm

PR #2312, opened 2022-05-23
Nt/editor toolbar spacing

noumantahir:
### What does this PR?

On android changed soft input mode from span to resize as it was make view add extra space between toolbar and keyboard. Also had to remove Keyboard avoiding view from android in editor as it was also affected the tool positioning.

It took a lot more time than what I anticipated in the start, yet the solution was very simple... that being said, the solution will also affect other places where keyboard is being used, so far no serious deviation is observed so we should be good to go.

### Issue number
https://github.com/orgs/ecency/projects/2#card-82286106

### Screenshots/Video

https://user-images.githubusercontent.com/6298342/169865373-7531ea2c-aff4-4c66-835c-536581f90cca.mov

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -31,7 +31,7 @@
         android:launchMode="singleTask"
         android:label="@string/app_name"
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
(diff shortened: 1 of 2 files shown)


noumantahir:
@feruzm anything that need addressing before we can merge this...?

feruzm (approved):
lgtm

PR #2314, opened 2022-05-24
Delegation Screen Issue

### LINKED ISSUE #2313: Unable set delegation amount
Bug: Delegation won’t work, On iPhone XS and my old 7 same…

One issue that might be bugging users is the fact there is not direct way to close the keypad.. try to check the return button to done button so users know how to close the keyboard.

Screenshot 2022-05-23 at 9 34 53 PM

Second issue I experience was the weired behaviour of input field if I add value greater than max....perhaps shifting post change calculations to onBlur might improve the experience

https://user-images.githubusercontent.com/6298342/169868274-4dca9cd0-4406-4688-85b9-4dea0ce9f786.mov

aliseyalvi:
### What does this PR?
In this PR

- added done button in keyboard so user can close keyboard when he is done with typing
- shifted input amount processing from realtime text change to onEndEditing i-e it will process and validate the max amount after keyboard is closed or done button is pressed. Meanwhile it also ensure that added amount is valid if user directly presses the review button
- added keyboard blur when review button is pressed which was causing a little glitch previously

### Issue number
fixes #2313 

### Screenshots/Video

https://user-images.githubusercontent.com/48380998/169959333-0df3c7a5-d59f-4cdc-8715-538c9d45bd7c.mp4

### DIFF
--- a/src/screens/transfer/screen/delegateScreen.js
+++ b/src/screens/transfer/screen/delegateScreen.js
@@ -217,7 +217,7 @@ class DelegateScreen extends Component {
     if (step === 1) {
       // this.setState({ step: 2 });
     } else {
-      // this.amountTextInput.current.blur();
+      this.amountTextInput.current.blur();
       let body =
         intl.formatMessage(
           { id: 'transfer.confirm_summary' },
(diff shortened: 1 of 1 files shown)


feruzm (approved):
lgtm

PR #2318, opened 2022-05-25
Fixes app starting in zoomed in state

### LINKED ISSUE #2310: App loads in zoomed in state..
happens if phone orientation was horizontal with devices's orientation locker disabled and ecency app is opened.

https://user-images.githubusercontent.com/6298342/169801791-ce18bab1-b290-49db-a3c0-aebd0eb3f5c0.mov

### LINKED ISSUE #2327: Lottie animations not responded to new dimensions strategy
This is while I was using app with landscape startup

https://user-images.githubusercontent.com/6298342/170863333-ffd950f9-82c0-4d03-9f5c-c05d08fecd81.mov

aliseyalvi:
### What does this PR?
This PR is trying to fix the zoomed in state when device is in landscape mode and app is started. 
We are locking our app to portrait but when device is in landscape it calculates dimensions based on landscape which causes dimension to set according to landscape but device is in portrait causing zoomed in behaviour.
Currently I have calculated dimensions based on device orientation , inverted them and applied on bottom tab component . And it has fixed the issue. We can extend this thing to whole app and replace the dimension being used in our app with our newly calculated dimensions. Later when we we'll add landscape support we can just change the dimensions in one place and it would reflect in all other places.

Unlock the device orientation from status bar and  tilt device to landscape.
Start ecency app 
When app will load you will see bottom tab component is now fixed and zoomed in issue is gone.
This issue still persist in other parts of the app which we can fix in this PR

### Issue number
fixes #2310 
fixes #2327 

### Screenshots/Video

https://user-images.githubusercontent.com/48380998/170174015-c4e49593-65ff-4b22-a330-17278ffdf81b.mp4


noumantahir:
@aliseyalvi have you tried `Dimensions.set()` to achieve app wise solution.... ?

noumantahir:
for reference:
https://github.com/react-native-toolkit/react-native-responsive-dimensions

src/utils/getWindowDimensions.ts
```diff
@@ -0,0 +1,23 @@
+import { Dimensions } from 'react-native';
+import {getOrientation, getInitialOrientation} from 'react-native-orientation-locker';
```
aliseyalvi, line 2: @noumantahir  this import is causing issue. These import do not exist. We have to import Orientation and then destructure it to get its methods

src/utils/getWindowDimensions.ts
```diff
@@ -0,0 +1,23 @@
...
+import { orientations } from '../redux/constants/orientationsConstants';
+
+
+const getWindowDimensions = () => {
+  const orientation = getOrientation ? getOrientation() : getInitialOrientation();
```
aliseyalvi, line 7: This getOrientation() method accepts a callback function and doesn't return anything. It has no use here because it is returning undefind here. For now getInitialOroentation would fulfil the purpose I think

aliseyalvi (commented):
@noumantahir there is a glitch in code which is preventing app building. I had fixed the import statement causing the isse and extend this method to other parts of the app where screen width is being calculated

aliseyalvi:
@noumantahir 

- apart from the above changes, replaced all occurences of Dimensions.get('window') and Dimensions.get('screen') with our utility method getWindowsDimensions() https://github.com/ecency/ecency-mobile/pull/2318/commits/79a60c555e24be4341ccd6ea8c4031f2c6002ed7
- There was a little glitch in tabBar components inside tabs when we swipe the tabs the underline border was getting wrong container widths. That is fixed in this commit https://github.com/ecency/ecency-mobile/pull/2318/commits/65a84c1938686e38b4e0e1b5c200e08600e6dd8d

Here is screen record which is captured in auto rotate lock off and mobile set to landscape and then Ecency is started. It should now calculated correct widths.

https://user-images.githubusercontent.com/48380998/170829308-eea4d545-6954-46f5-a1cd-06d4d24f83be.mp4

noumantahir (commented):
@aliseyalvi any reason for not removing this usage of dimensions...?

https://github.com/ecency/ecency-mobile/blob/420827b6c4381a1892cc701ee57246c93619a573/src/components/tabBar/view/tabBarView.js#L67

aliseyalvi:
> @aliseyalvi any reason for not removing this usage of dimensions...?
> 
> https://github.com/ecency/ecency-mobile/blob/420827b6c4381a1892cc701ee57246c93619a573/src/components/tabBar/view/tabBarView.js#L67

Actually the TabBar component was calculating wrong dimensions for its underline. Dont know the exact reason why only Tabbar component was getting the different Dimensions. So thats why I had kept it same and it worked fine. Here u can see it in this commit https://github.com/ecency/ecency-mobile/pull/2318/commits/65a84c1938686e38b4e0e1b5c200e08600e6dd8d

aliseyalvi:
> > @aliseyalvi any reason for not removing this usage of dimensions...?
> > https://github.com/ecency/ecency-mobile/blob/420827b6c4381a1892cc701ee57246c93619a573/src/components/tabBar/view/tabBarView.js#L67
> 
> Actually the TabBar component was calculating wrong dimensions for its underline. Dont know the exact reason why only Tabbar component was getting the different Dimensions. So thats why I had kept it same and it worked fine. Here u can see it in this commit [65a84c1](https://github.com/ecency/ecency-mobile/commit/65a84c1938686e38b4e0e1b5c200e08600e6dd8d)

noumantahir:
> > > @aliseyalvi any reason for not removing this usage of dimensions...?
> > > https://github.com/ecency/ecency-mobile/blob/420827b6c4381a1892cc701ee57246c93619a573/src/components/tabBar/view/tabBarView.js#L67
> > 
> > 
> > Actually the TabBar component was calculating wrong dimensions for its underline. Dont know the exact reason why only Tabbar component was getting the different Dimensions. So thats why I had kept it same and it worked fine. Here u can see it in this commit [65a84c1](https://github.com/ecency/ecency-mobile/commit/65a84c1938686e38b4e0e1b5c200e08600e6dd8d)
> 
> This is also explained in this comment [#2318 (comment)](https://github.com/ecency/ecency-mobile/pull/2318#issuecomment-1140273969)

@aliseyalvi  Alright, I have updated getWindowOrientations to support this as nativeWidth and nativeHeight to avoid confusions in future.

noumantahir (approved):
@feruzm all good from my side now....you may review

Issue #2311, opened 2022-05-23
assess modal closure while adding a comment


noumantahir:
**My Assessments...**

It's possible that while adding comment using tries to set cursor and that accidentally closes keyboard double tap closure modal in a hurry.. like the video below.

https://user-images.githubusercontent.com/6298342/170041335-6707652e-0b1c-47d6-bef1-c281331b572d.mov

noumantahir:
My attempts so far.

- tried to play with keyboard related props actions sheet,
- tried using react-native-modalize, same issues faced, though the keyboard appearance transition is much better 

seems there is no way in ios to avoid keyboard dismiss unless we create our own small UI for modal, just like we did with toastNotifications

aliseyalvi:
@noumantahir here is some workaround which I found on issues page of react-native-portalize. Seems like the cursor jumping inside portal is a known issue. That is due to state update in same component where portal is being called. As a workaround we can create a HOC for text input. Please take a look https://github.com/jeremybarbet/react-native-portalize/issues/6#issuecomment-673299898

noumantahir:
Marked as Done in Project board already

PR #2328, opened 2022-05-29
Nt/reply modal tweaks

noumantahir:
### What does this PR?
Minor tweaks that should improve quick reply, please do give it a shot as well. Will be redoing it as well after we are done with critical issues we are facing.

PR merges another relevant branch from @aliseyalvi as well.

### Issue number
#2311

### Screenshots/Video
Screenshot 2022-05-29 at 3 57 59 PM


aliseyalvi:
@noumantahir 

- I have added the content of quick reply modal into child component and put the portal back as Higher Order Component which solves the initial cursor jumping issue due to which we have removed the portal but that raised Keyboard Taps Persistance issue. This would solve both issues . check this commit https://github.com/ecency/ecency-mobile/pull/2328/commits/a45ed69017aa1b1a05cd96fa7c47997b99deca50
- in this commit added check for navigation params which was cuasing error when comment is done in full editor when naigating from quick reply to full editor through expand btn https://github.com/ecency/ecency-mobile/pull/2328/commits/83cf0f89ac8595cb50f784b12e53eacc9f4824fa

noumantahir:
@aliseyalvi apparently keyboard is not auto opening in android... also try if you can set separate bottom paddings for keyboard opened and close...

aliseyalvi:
> is keyboard not opening in quick reply modal or when navigating to full editor for reply?

@noumantahir I had tried to update the bottom padding based on keyboard opening and closong but it is showing a weired behaviour. I mean this is not smooth and shows a jerk when changing padding. See the screen record 

https://user-images.githubusercontent.com/48380998/171040884-ba33e285-4004-48f3-ae1b-ac20b0e89186.mp4

noumantahir:
> > @aliseyalvi apparently keyboard is not auto opening in android... also try if you can set separate bottom paddings for keyboard opened and close...
> 
> is keyboard not opening in quick reply modal or when navigating to full editor for reply? I had tested on my device and keyboard is opening on both quick reply modal and full editor as well

perhaps it was emulator issue, tested on my phone instead of emulator, working fine

noumantahir:
@feruzm this can be tested and merged now... the primary change is now user do not have to close modal before approving reply while the input glitch is no longer affecting us.

feruzm (approved):
Lgtm, let’s see if Portal doesn’t cause issue on iPhone again

PR #2332, opened 2022-05-30
Validate ecency url in boost screen

aliseyalvi:
### What does this PR?
This PR fixes the issue of not validating ecency url in boost screen. 
Previously boostscreen only accepts permlink selected from autocomplete dropdown and when coming directly from a post. Now it accepts valid ececny url and formats ecency url into author/permlink.

### Screenshots/Video

Uploading screencast 2022-05-30 22-27-11.mp4…


noumantahir (approved):
Tested with various different urls both ecency internal and external, seems to work well in my tests

PR #2340, opened 2022-06-07
Sa/quick comment cache

### LINKED ISSUE #2339: maintain cache of quick reply modal
Also, can't paste any copied text in quick reply.

aliseyalvi:
### What does this PR?
This PR adds cache in quick reply modal. It maintains separate cache for each post.
It saves comment value in cache when quick reply modal is closed and deletes the cached entry when comment is successfully done or expiry time is passed

### Issue number
fixed #2339 

### Screenshots/Video

Uploading screencast 2022-06-07 11-03-11.mp4…


src/components/quickReplyModal/quickReplyModalContent.tsx
```diff
@@ -8,43 +8,85 @@ import { useSelector, useDispatch } from 'react-redux';
...
+  const _addQuickCommentIntoCache = () => {
+    const date = new Date();
+    const updatedStamp = date.toISOString().substring(0, 19);
+
+    const quickCommentCacheData = {
+      parent_permlink: parentPermlink,
+      body: commentValue,
+      created: quickCommentCache ? quickCommentCache.created : date ,
+      updated: updatedStamp,
+      expiresAt: date.getTime() + 6000000,
```
noumantahir, line 84: @aliseyalvi @feruzm for how long do you think we should keep the reply/comment drafts saved... 7 Days should be good I guess

src/redux/store/store.ts
```diff
@@ -11,12 +11,14 @@ const transformCacheVoteMap = createTransform(
...
-    comments : Array.from(inboundState.comments)
+    comments : Array.from(inboundState.comments),
+    quickComments : Array.from(inboundState.quickComments),
   }),
   (outboundState) => ({ 
     ...outboundState, 
     votes:new Map(outboundState.votes),
-    comments:new Map(outboundState.comments)
+    comments:new Map(outboundState.comments),
+    quickComments: new Map(outboundState.quickComments)
```
noumantahir, line 21: instead of `quickComments`, name them `drafts`.... moving forward this object will also maintain post drafts as well, it will also avoid confusion as well....

src/components/quickReplyModal/quickReplyModalContent.tsx
```diff
@@ -8,43 +8,85 @@ import { useSelector, useDispatch } from 'react-redux';
...
+    _addQuickCommentIntoCache();
+  };
+
+  // add quick comment value into cache
+  const _addQuickCommentIntoCache = () => {
+    const date = new Date();
+    const updatedStamp = date.toISOString().substring(0, 19);
+
+    const quickCommentCacheData = {
+      parent_permlink: parentPermlink,
```
noumantahir, line 80: instead of using `parent_permlink`, use simple prop `id` with value `${parentAuthor}/${parentPermlink}` .... this will simplify it's scalability with post drafts cache... check `setDraftPost` method in `realm.js`

src/redux/reducers/cacheReducer.ts
```diff
@@ -26,19 +26,29 @@ export interface Comment {
     expiresAt?:number,
 }
 
+export interface QuickComment {
```
noumantahir, line 29: do also keep track to `username` making the comment... I made a draft using @noumantahir but it was also visible when using @demo.com.

for reference you can check `getDraftPost` and `setDraftPost` method in `realm.js` but no need to follow the same logic.

Screenshot 2022-06-07 at 12 52 01 PM

noumantahir (changes requested):
@aliseyalvi added few improvement suggestion and one bug in review comments

src/components/quickReplyModal/quickReplyModalContent.tsx
noumantahir: @aliseyalvi make it 7 days expiration time

aliseyalvi:
@noumantahir 

- renamed quickComments to drafts in redux
- added author in drafts data and username with draftId . Actually as we were using ${parentAuthor}/${parentPermlink} as key for Map which could be same for other account as well. I had made it unique by concating username e-g `${currentAccount.name}/${parentAuthor}/${parentPermlink}`. It should solve this issue https://github.com/ecency/ecency-mobile/pull/2340#discussion_r890885717

src/components/quickReplyModal/quickReplyModalContent.tsx
noumantahir: @aliseyalvi `parent_permlink` is not serving any purpose in current arrangement, perhaps can be removed to avoid confusion

aliseyalvi: updated

aliseyalvi:
@noumantahir  updated https://github.com/ecency/ecency-mobile/pull/2340/commits/b09192c2b53c45aa3d03e461078c530fcb328261

noumantahir (approved):
lgtm... this PR solved problem for quick comment cache... there will be another phase to handle extension of that cache in primary editor

PR #2343, opened 2022-06-07
Fixed 3speak videos don't play in fullscreen

### LINKED ISSUE #2342: Unable to zoom in 3Speak video to full screen in android.
reference post: https://ecency.com/hive-181335/@unklebonehead/eoabdyyv
This is a known issue as we are playing 3speak and other embedded videos in webview and webview is causing some issue in triggering fullscreen espacially on android. on iOS side it is working fine

aliseyalvi:
### What does this PR?
This PR fixes #2342 issue in which 3speak videos do not go to fullscreen after tapping fullscreen. This bug was happening on android side as webview was causing this issue.
react-native-webview offers a prop allowsFullscreenVideo which activates the fullscreen feature but that leads to another bug. The 3speak videos after going to fullscreen and coming back to normal it opens in picture in picture which causes a weired experience as it got stuck at some random position.

The solution I had implemented is that instead of passin uri to webview I had passed uri to iframe in html in webview which solves that glitch.

### Issue number
fixes #2342 

### Screenshots/Video

https://user-images.githubusercontent.com/48380998/172441884-bc3e235b-44c7-44bf-9a1c-9f507683aeb4.mp4


noumantahir (changes requested):
the video goes to full screen but if device's orientation is locked, the fullscreen video does not support landscape mode... another thing I noticed is, the setIsFullScreen state is only linked with youtube... 

at-least the video should go horizontal fullscreen even if device orientation is locked..for confirmation I tried 3Speak video playback in mobile browser and it supported landscape fullscreen  even when device orientation is locked.

We will be working a lot in support of video posts in coming future, so it's better make video playback experience smooth as possible before we get to that.

aliseyalvi:
> the video goes to full screen but if device's orientation is locked, the fullscreen video does not support landscape mode... another thing I noticed is, the setIsFullScreen state is only linked with youtube...
> 
> at-least the video should go horizontal fullscreen even if device orientation is locked..for confirmation I tried 3Speak video playback in mobile browser and it supported landscape fullscreen even when device orientation is locked.
> 
> We will be working a lot in support of video posts in coming future, so it's better make video playback experience smooth as possible before we get to that.

@noumantahir actually there is no direct way to detect wether video is in fullscreen or not in react-native-webview so as we are locking the orientation and unlock only when isFullScreen is true, so in the case of webview we can't unlock the orientation lock

noumantahir (approved):
this can be merged for it solves one of the problem under discussion..

feruzm (approved):
lgtm

PR #2361, opened 2022-06-22
[WIP] Caching in Communities 

### LINKED ISSUE #2351: Use cached subscribedCommunities in communities screen to avoid getSubscriptions re-fetch
Subscribed communities data is already being fetched and stored in redux, all we have to do now is used that same store data in places where user communities is listed... 

`(state) => state.communities.subscribedCommunities`

- [ ] 1. in communities screen accessed from side menu

- [ ] 2. in editor while selecting community

- [ ] Another thing we need to take care for is to update cache every time user join or leave a community, update cache when user opens community selection for example in editor or communities screen. 

- [ ] And remove community cache refresh on app startup.

Screenshot 2022-06-11 at 1 10 52 PM
Screenshot 2022-06-11 at 1 15 14 PM

aliseyalvi:
### What does this PR?
This PR is WIP for adding cache support in communities screen and using cached data wherever communities data in being used in app.

Note: There is certain glitch which is after leaving community it instantly updates on membership tab but if we switch to discover tab immediatly we will see same community is subscribed, this is due to delayed refetch, after refetching data if status becomes correct. The local cache has to be maintained in both tabs and vise versa
### Issue number
fixes #2351 

### Screenshots/Video

https://user-images.githubusercontent.com/48380998/175119819-c7755ddd-951b-4258-a824-37556c5649f2.mp4


src/redux/reducers/communitiesReducer.js
```diff
@@ -110,6 +111,24 @@ export default function (state = initialState, action) {
           error: action.payload,
         },
       };
+    case UPDATE_SUBSCRIBED_COMMUNITIES:
+      let itemIndex = state.subscribedCommunities.data.findIndex(
```
noumantahir, line 115: instead of manually updating the list, perhaps we should make use of cache store and keep data well organised...

src/screens/communities/container/communitiesContainer.js
```diff
@@ -26,88 +30,90 @@ const CommunitiesContainer = ({ children, navigation }) => {
...
-            }),
-          );
-
-          setSubscriptions(subs);
-          setDiscovers(shuffle(communities));
-          setIsSubscriptionsLoading(false);
+    if (subscribedCommunities && subscribedCommunities.length > 0) {
+      setIsSubscriptionsLoading(true);
+      let subs = subscribedCommunities.slice();
+      getCommunities('', 50, null, 'rank').then((communities) => {
```
noumantahir, line 98: This is probably causing a loading glitch when first entering communities screen, perhaps we should keep the two tabs data separate... show subscribed communities from redux cache separate, and fetch + post process ranked communities separate

noumantahir (changes requested):
I think the overall implementation can use some cleaning and better data organising. add a couple of comments in code, feel free to reach out for any confusions/clarifications etc.

PR #2362, opened 2022-06-23
Nt/register device

noumantahir:
### What does this PR?
push token api migrated to register-device, there was a little trick to make it work for multiple logged in users.

What really surprised me however is the fact the access token decrypted using default pin worked well even when custom pin was set even though to my calculation it was not supposed to work. So either I was overthinking or there must be something off in updating app pin which we might want to dig more later.

### Screenshots/Video
Screenshot 2022-06-24 at 4 11 47 AM

### DIFF
--- a/src/providers/ecency/ecency.ts
+++ b/src/providers/ecency/ecency.ts
@@ -450,11 +450,15 @@ export const markNotifications = async (id: string | null = null) => {
 };
 
 
-export const setPushToken = async (data) => {
+export const setPushToken = async (data, accessToken = null) => {
   try {
     if (!data.username) {
       console.log("skipping push token setting, as no user is provided")
       return;
+    }
+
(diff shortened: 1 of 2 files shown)


feruzm (approved):
We should for sure check why pin is not encrypting keys. So it is only making app inaccessible that’s it?! If we fix that, should make sure registering device is still functional which might be tricky indeed without user pin.
@noumantahir

noumantahir:
@feruzm it is re-encrypting keys but only for logged in user... keys in `otherAccounts` section of redux are not changed

feruzm:
@noumantahir so active user’s keys are encrypted with user pin and rest is encrypted with default pin? In that case maybe this is still good because otherwise we would need to get user pin if we want them to update fcm token or register device. Current situation seem good now

PR #2373, opened 2022-07-18
Nt/quick reply tweaks

noumantahir:
### What does this PR?
Re-wired and tested functionality of quick reply modal.... 

**Improvements**
- Only one instance of reply modal is used instead of multiple
- using native props to update textinput values for improved performance
- There was an unnecessary hook getting called on every text update, removed that for a more clean draft update implementation. 
- Improved cache update of modal closure


feruzm (approved):
lgtm

PR #2375, opened 2022-07-20
adjusted contents of submitReply

noumantahir:
### What does this PR?
Tested commenting 60 times after adding this fix... if app crashes now, I doubt it will be because of comments.


feruzm (approved):
lgtm

PR #2376, opened 2022-07-21
Nt/editing tweaks

noumantahir:
### What does this PR?
This PR removes save to storage as user write feature and saves local drafts when editor is closed... It is indeed a downgrade we should test if this stabilises the app and fixes the crash problem and next try to find a better approach to add save on the go functionality.

I have test this again and again and even tested through automated script for upto 5000 comments in a single app session... no memory problem, no crash showed up. Let's see how things go for melinda

If app still crashes as before for melinda, perhaps we can safely put back the current save on the go functions as well, and look for some other source for crash... but I am quite certain this will fix the issue...


feruzm (approved):
let's test

PR #2380, opened 2022-07-23
Bug Fixed: Header not hiding in login screen when keyboard is active

### LINKED ISSUE #2379: Fix login screen header hiding inputs in android.. must test on nexus 5 emulator

aliseyalvi:
…ctive

### What does this PR?
This PR fixes a bug in login screen which was not hiding header when keyboard is active. The reason behind this bug was onKeyboardWillShow event in react-native-keyboard-aware-scroll-view which was actually triggering keyboard update. But this event method is not supported on android so android was not hiding the header.
The simple fix was instead of using the above event method I had updated the keyboard active status based on input focus/blur.

### Issue number
fixes #2379 

### Screenshots/Video
Tested on iphone 11 simulator, Poco M4 Pro android 11 and Nexus 5 simulator. Nexus 5 simulato screen recording is attached.

https://user-images.githubusercontent.com/48380998/180616108-3c69a223-6d76-4b33-8fbd-df40cc164193.mp4

### DIFF
--- a/src/components/formInput/view/formInputView.tsx
+++ b/src/components/formInput/view/formInputView.tsx
@@ -26,6 +26,7 @@ interface Props extends TextInputProps {
   inputStyle:TextStyle;
   isValid:boolean;
   onChange?:(value:string)=>void;
+  handleFocus?:(value:boolean)=>void;
 }
 
 const FormInputView = ({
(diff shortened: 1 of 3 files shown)


feruzm (approved):
lgtm 👍

PR #2382, opened 2022-07-24
Bug Fix: native Image sharing not working

### LINKED ISSUE #2381: Native Image sharing opens app but doesn't upload image, Text sharing seem to be working.

aliseyalvi:
### What does this PR?
This PR fixes issue of native image/text not sharing to ececny app. When navigating to editor it wasn't passing the received parameters correctly. Fixed the issue by catching received file in applicationContainer and passed into editor as navigation parameters.

Note:  There is an issue that is it shows image upload error sometime when sharing image. I think that is because our image upload module is not stable enough and that  is giving upload error. May be we can check that issue in separate PR or issue we have already created for remaking of upload process

### Issue number
fixes #2381 

### Screenshots/Video

https://user-images.githubusercontent.com/48380998/180657419-37607429-68e3-4d10-a0d2-7121852aa811.mp4

### DIFF
--- a/src/screens/application/container/applicationContainer.tsx
+++ b/src/screens/application/container/applicationContainer.tsx
@@ -144,13 +144,14 @@ class ApplicationContainer extends Component {
     this._fetchApp();
 
     ReceiveSharingIntent.getReceivedFiles(
-      () => {
+      files => {
         navigate({
           routeName: ROUTES.SCREENS.EDITOR,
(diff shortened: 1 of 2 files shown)


feruzm (approved):
lgtm, @aliseyalvi let's try on separate PR to improve that popup when shows after image is selected for share, for Text you get input filled up but for file/image, input seem odd, maybe we can change it or use something different?

PR #2387, opened 2022-07-26
Nt/biometric unlock

noumantahir:
### What does this PR?
App now support unlocking using biometric... fully tested on ios, partially tested on android, will be testing on android shortly.

Update: tested on android as well

- Setting new pin or changing ping ignores biometric and requires old pin input.
- All other processes requires biometric if enabled and enrolled

### Issue number
https://github.com/orgs/ecency/projects/2#card-83764380

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/180994108-99e86323-f78e-430a-ae8d-4747b9022e38.mov


feruzm (approved):
Lgtm, only would be better if we could offer unlock same way as pin, not after pin

PR #2405, opened 2022-07-30
Bug Fix: Quick reply not opening in profile comments tab

### LINKED ISSUE #2404: Bug: Reply button not working on feed screen
when I try to reply to comment from my profile in mobile app I touch reply but nothing happens, I need to go to post to reply to the comment. It's not big issue, no problem but just to know. Here's the screenshot

ref:
https://discord.com/channels/385034494555455488/385448634197737484/1002597853870436412

aliseyalvi:
### What does this PR?
Changed implementation of onReplyPress method and handled it inside commentView and handle showReplyModal dispatch action in commentView.
Note: As now we are doing comments from quick reply then there is no need to handle onReplyPress in different components instead we can handle it in commentView.tsx / comment module and open quick reply modal here. This will reduce the complexity caused by onReplyPress passing from many layers and also reduces the bug caused by missing onReplyPress prop which actually was the root cause in the repective issue

### Issue number
fixes #2404


noumantahir (commented):
Ideally we should not use redux in a view based class in accordance to previous structure implementation but apparently it is fixing the issue so we can probably rely on that....

noumantahir (approved):

feruzm (approved):

PR #2417, opened 2022-08-11
Nt/navigation

noumantahir:
### What does this PR?
Upgraded navigation to version 5.0

This part one, in second stage will be upgrading to v6.0 if we do not achieve the modal performance we expect from this upgrade. Though upgrade from 5 to 6 is a lot straight forward as it utilises same navigation structure unlike 4 to 5 upgrade

**Updates 30 August**
- We are now at react navigation 6
- Now using native stack instead of js stack
- pincode and welcome modals are now navigable screens to avoid unnecessary global modals
- Support for nested stack with login, signup, pincode and welcome screens part of root stack

### Issue number
https://github.com/orgs/ecency/projects/2#card-84787829
https://github.com/orgs/ecency/projects/2#card-84579161 (This also got covered)

**_PR requires rigorous testing_**


noumantahir:
Update: now that we will be including this in next release, upgrading to v6 navigation, as it will utilise all native based navigation including modals that should give a good performance optimisation to app

src/containers/pointsContainer.js
```diff
@@ -100,12 +99,13 @@ const PointsContainer = ({
...
 
     if (isPinCodeOpen) {
-      dispatch(
-        openPinCodeModal({
-          navigateTo,
-          navigateParams,
-        }),
-      );
+      navigation.navigate({
+        routeName: ROUTES.SCREENS.PINCODE,
```
feruzm, line 103: Route name changed to name right?

src/containers/walletContainer.js
```diff
@@ -255,12 +252,13 @@ const WalletContainer = ({
     }
 
     if (isPinCodeOpen) {
-      dispatch(
-        openPinCodeModal({
+      navigate({
```
feruzm, line 255: Here also

src/screens/application/container/applicationContainer.tsx
```diff
@@ -338,6 +346,17 @@ class ApplicationContainer extends Component {
     dispatch(purgeExpiredCache());
   };
 
+  _startPinCodeTimer = () => {
+    const {isPinCodeOpen:_isPinCodeOpen} = this.props;
+    if (_isPinCodeOpen) {
+        this._pinCodeTimer = setTimeout(() => {
+            navigate({
+              routeName:ROUTES.SCREENS.PINCODE
```
feruzm, line 354: Here also?!

src/screens/coinDetails/screen/coinDetailsScreen.tsx
```diff
@@ -130,12 +129,13 @@ const CoinDetailsScreen = ({ navigation }: CoinDetailsScreenProps) => {
     }
 
     if (isPinCodeOpen) {
-      dispatch(
-        openPinCodeModal({
+      navigate({
+        routeName: ROUTES.SCREENS.PINCODE,
```
feruzm, line 133: Here?

src/screens/coinDetails/screen/coinDetailsScreen.tsx
```diff
@@ -130,12 +129,13 @@ const CoinDetailsScreen = ({ navigation }: CoinDetailsScreenProps) => {
...
+        params:{
           navigateTo,
           navigateParams,
-        }),
-      );
+        }
+      })
     } else {
       navigate({
         routeName: navigateTo,
```
feruzm, line 141: Here

src/screens/steem-connect/hiveSigner.js
```diff
@@ -51,12 +51,13 @@ class HiveSigner extends PureComponent {
               dispatch(loginAction(true));
 
               if (isPinCodeOpen) {
-                dispatch(
-                  openPinCodeModal({
+                navigation.navigate({
+                  routeName: ROUTES.SCREENS.PINCODE,
```
feruzm, line 55: Here?

feruzm (approved):

src/containers/pointsContainer.js
noumantahir: it is changed yes, but in most of our application we were using routeName, to keep the PR simple I am using the compat package provided by new navigation

noumantahir: a lot of files in PR uses just this change, that helped keep using navigation the way we were using before. later, slowly we can migrate away from compat as well
Screenshot 2022-09-08 at 4 28 41 PM

src/screens/coinDetails/screen/coinDetailsScreen.tsx
noumantahir: The navigate service is also utilising compat package, that is why routeName is kept here as well.
Screenshot 2022-09-08 at 4 31 54 PM

src/screens/steem-connect/hiveSigner.js
noumantahir: pinCode is now a navigable screen instead of modal, that is why no longer need to dispatch

src/containers/pointsContainer.js
feruzm: Understood 👍

PR #2421, opened 2022-08-12
update estm value fetch

noumantahir:
- Updated market-data/latest to process estm
- migrate market-data/currency-rate to private-api end point


feruzm (approved):
lgtm

PR #2434, opened 2022-08-26
[WIP] working on mute/unmute in feed

aliseyalvi:

### DIFF
--- a/src/components/postDropdown/container/postDropdownContainer.tsx
+++ b/src/components/postDropdown/container/postDropdownContainer.tsx
@@ -298,8 +298,11 @@ class PostDropdownContainer extends PureComponent {
   };
 
   _addToBookmarks = () => {
-    const { content, dispatch, intl } = this.props as any;
-
+    const { content, dispatch, intl, isLoggedIn, navigation } = this.props as any;
+    if(!isLoggedIn){
+      showLoginAlert({navigation, intl});
+      return;
+    }
     addBookmark(get(content, 'author'), get(content, 'permlink'))
       .then(() => {
         dispatch(
@@ -322,7 +325,11 @@ class PostDropdownContainer extends PureComponent {
   };
 
   _reblog = () => {
-    const { content, currentAccount, dispatch, intl, isLoggedIn, pinCode } = this.props as any;
+    const { content, currentAccount, dispatch, intl, isLoggedIn, pinCode, navigation } = this.props as any;
+    if(!isLoggedIn){
+      showLoginAlert({navigation, intl});
+      return;
+    }
     if (isLoggedIn) {
       reblog(currentAccount, pinCode, content.author, get(content, 'permlink', ''))
         .then(() => {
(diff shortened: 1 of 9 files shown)


src/components/postDropdown/container/postDropdownContainer.tsx
```diff
@@ -177,11 +182,137 @@ class PostDropdownContainer extends PureComponent {
...
+          ),
+        );
+      })
+      .catch((err) => {
+        this._profileActionDone({ error: err });
+      });
+ 
+  };
+
+  _handleFollowUnfollowUser = async () => {
```
noumantahir, line 245: @aliseyalvi do we really need this considering, I think we can just keep feed limited to blocking. user my unblock later from profile, also we can give option somewhere in profile to view a list of blocked users and unblock them if required.

aliseyalvi: I think if post is showing in screen but content is hidden, it would be easy for user to unblock again. But we can remove that , what u suggest

src/components/postDropdown/container/postDropdownContainer.tsx
```diff
@@ -177,11 +182,137 @@ class PostDropdownContainer extends PureComponent {
...
+      following,
+    })
+      .then(() => {
+        const curMutes = currentAccount.mutes || [];
+        if (curMutes.indexOf(username) < 0) {
+          //check to avoid double entry corner case
+          currentAccount.mutes = [username, ...curMutes];
+        }
+        dispatch(updateCurrentAccount(currentAccount));
+        onLoadPosts(true); //fetch posts again after muting user
```
noumantahir, line 230: instead of passing this prop from grad parent, perhaps better option would to hive content on feed, I think there is an option for that... the card can keep tracked of muted blocked users from currentAccount and hide content using that instead passing a prop deep inside.

noumantahir (commented):
In summary, I believe 

1. we do not need to pass onLoadPosts to grant parents and can user currentAccount data change event to hide content.
2. Perhaps we can limit dropdown option to Block only, no need to unblocking option in feed screen

noumantahir:
Needs to handle logged out state.

Screenshot 2022-08-27 at 12 23 53 PM

src/components/postDropdown/container/postDropdownContainer.tsx
```diff
@@ -177,11 +182,137 @@ class PostDropdownContainer extends PureComponent {
...
+      case 'unmute':
+        this._handleMuteUnmuteUser(!isMuted);
+        break;
       default:
         break;
     }
   };
 
+  _handleMuteUnmuteUser = async (isMuteAction) => {
+    const { currentAccount, content } = this.props as any;
```
noumantahir: Needs to handle logged out state somewhere..

Screenshot 2022-08-27 at 12 23 53 PM

noumantahir (commented):
Needs to handle logged out state. Other than that I think it's working perfectly

Screenshot 2022-08-27 at 12 23 53 PM

feruzm (approved):
Lgtm

PR #2436, opened 2022-08-28
sa/delete account

aliseyalvi:
### What does this PR?
Added delete account button in settings screen.
Upon delete account, user is logged out and request is sent on endpoint for account deletion
Also fixed a small issue in quick reply overlay background
### Screenshots/Video

https://user-images.githubusercontent.com/48380998/187090035-eafee04e-8a6a-4e79-97cd-6f0728d4181b.mp4


feruzm (approved):
Lgtm

PR #2443, opened 2022-09-01
nt / malformed local reporting

noumantahir:
### What does this PR?
This is an attempt to track/report users with nullified/malformed local data stored

@feruzm apparently, the [issue](https://app.bugsnag.com/ecency/ecency-mobile/errors/627ced64abea150008359a54?filters[error.status]=for_review&filters[event.since]=30d&filters[app.release_stage]=production&sort=events) is indeed coming from previous app version, but we can still push this update to track down if users have malformed local data saved. this might be helpful but this can also wait till next month or so, by that time we will know if this issue is really persisting


feruzm (approved):
Yes it will go out with September release later

Issue #2430, opened 2022-08-19
Show upload progress modal while uploading images in editor during  post creation process


aliseyalvi:
Suggestion From Nouman: Add image thumbs as placeholder in images modal with activity indicators/progress

- [x] add basic upload function
- [x] update UI
- [x] update and test new selection and upload methods
- [x] fixed isUploading state causing glitch during multiple uploads
- [x] allow inserting images without losing editor focus and keyboard
- [x] include delete feature
- [x] test deep link media insertion

noumantahir:
Current implementation of media insertion is more intrusive than it has to be... below is my suggestion to improve that behaviour. a media section that would appear on media insert button click, but will not obstruct user's flow on editing post, user can easily change input points and insert images with single tap

![Screenshot 2022-09-01 at 4 44 35 PM](https://user-images.githubusercontent.com/6298342/187907207-7b435b34-4676-490d-bcbb-be36b4a44c37.png)

PR #2448, opened 2022-09-07
Nt/media insertion

noumantahir:
### What does this PR?

- Gallery and camera directly inserts item in content. but unlike web version, I am not adding image to our uploaded images
- Using + (Add) button user can add reusable uploaded
- Using - (Minus) enables delete mode for removing items from uploaded content
- Enabled sharing images from gallery app in active editor

The feature for adding image directly in content before uploaded finishes was very tricky to implement with out `setNativeProps` editor flow as it was causing some glitches when uploads finishes while user is writing content. Most of time is consumed to add a logic to avoid such glitches and it has to be tested well.

### Issue number
https://github.com/ecency/ecency-mobile/issues/2430

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/189109864-6f7b383d-8d6e-4df5-987c-e66ae76dd950.mov

https://user-images.githubusercontent.com/6298342/189109894-f9f75f24-5527-40ad-b89c-99afdf919631.mov


feruzm (approved):
Lgtm

PR #2452, opened 2022-09-09
Bug Fix: Autocorrect/Capital in tags input causes input hide

### LINKED ISSUE #2451: Improve: Change keyboard type to not auto correct tags inputs
Report: When my phone autocorrects  Chicago to a capital letter, I get the error message, but couldn't figure out a way  to go back and correct it, other than exiting and going back in.

ref: https://discord.com/channels/@me/920267778190086205/1017545389202739322

aliseyalvi:
### What does this PR?
This PR fixes issue of tag input get hidden when there is a capital letter
Setting fixed height to text input fixes the issue

### Issue number
fixes #2451


noumantahir (changes requested):
are you sure this is the solution to melinda's problem, because that bug is related to keyboard type capitalising word, please check again. Also, melinda is not using android oreo

aliseyalvi:
> are you sure this is the solution to melinda's problem, because that bug is related to keyboard type capitalising word, please check again. Also, melinda is not using android oreo

@noumantahir I think it should fix the issue as I was able to reproduce it. It was caused by changing of height of tag input when warning is visible. I had set the fixed height of tags Input instead of its container so its height shouldn't be changed when warning will be shown. U can merge it and Let melinda test it again. The android oreo condition was already there so I didn't changed that. I just changed the places which was actually causing issue

noumantahir:
@aliseyalvi this is the issue faced by melinda, for instance if I type pakistan with small caps, the auto correct would change that to pakistan with capital 'P'... I want you to do two things, 

1. disable auto-correct for tags input, it may or may not work, but try..
2. onTextChange convert text to app small caps to add redundency to the process. so even if I type capital 'Pakistan' intentionally, the post processor will convert it to 'pakistan'.s
 
Screenshot 2022-09-10 at 12 32 30 PM

noumantahir (changes requested):
Working is good, only one small issue, visible-password is only available in android, that is throwing error on ios.

Screenshot 2022-09-13 at 9 17 44 AM

noumantahir (approved):

feruzm (approved):
Lgtm

PR #2467, opened 2022-09-15
Nt/auto login

noumantahir:
### What does this PR?
Tested auto login pr after development merge and fixing navigation bug


feruzm (approved):
oh right, 👍

PR #2468, opened 2022-09-15
changed walletLineItem icon background to use variable color

aliseyalvi:
### What does this PR?
Changed hard coded icon background color to variable colors used in theme file. It fixes the issue of icons not showing due to same colors of background and icon in dark mode

### Screenshots/Video
image
image


noumantahir (approved):
Seems good

feruzm (approved):

PR #2472, opened 2022-09-16
Bug Fix: getSize error in drafts list

### LINKED ISSUE #2471: Bug: getSize error while loading drafts

aliseyalvi:
### What does this PR?
Replaced getSize method of react-native-image-size with method provided by Image module from react-native
Removed this old package react-native-image-size which was updated before 3 years ago which was causing error

### Issue number
fixes #2471 

### Screenshots/Video

https://user-images.githubusercontent.com/48380998/190665629-459df488-d41f-4546-829e-d7ad6933cb91.mp4


noumantahir (changes requested):
@aliseyalvi the reason why react-native-image-size was used in the first place is because some android devices were not returning right image size 
https://github.com/facebook/react-native/issues/33498

that being said, both Image.getSize and ImageSize.getSize are bypassed by `FastImage's` `onLoad` implementation in postCardView.js
https://github.com/ecency/ecency-mobile/blob/development/src/components/postCard/view/postCardView.js#L155

We can either make use to that approach, or simply avoid calculation image calculation for drafts list altogether as there is not much use for it anyways, we should be fine fixing thumb height for draft list item.

noumantahir (approved):
@aliseyalvi  all good, just try to avoid inline styles wherever possible, in this case we should have easily avoid it...

noumantahir:
Update: using FastImage instead of Progressive Image

feruzm (approved):
👍

PR #2473, opened 2022-09-17
limit images to 10 while extracting meta

### LINKED ISSUE #2460: Limit number of images to 10 in meta
Also remember if user adds 100 pictures in post, we should only include first 10 image links into json metadata, all not necessary extra bytes/bits Also normal links

aliseyalvi:
### What does this PR?
Filtered only first 10 images extracted from post body while creating post meta data

### Issue number
fixes #2460


src/utils/editor.ts
```diff
@@ -222,7 +222,7 @@ export const extractMetadata = (body:string, thumbIndex?:number) => {
   const mUrls = extractUrls(body);
   const mUsers = body && body.match(userReg);
 
-  const matchedImages = extractImageUrls({urls:mUrls});
+  const matchedImages = extractImageUrls({urls:mUrls}).filter((element,index) => index < 10); //filter only first 10 images
```
noumantahir, line 225: @aliseyalvi please test all use-cases of returned object `matchedImages` and test varying use cases of extraMetaData, I have a feeling it will be disturb some app functions

noumantahir (changes requested):

src/utils/editor.ts
aliseyalvi: @noumantahir apparently it is working well. Also tested thumb selection process and it is working fine as well.

noumantahir:
@aliseyalvi this should lead you to the issue I have been referring to, please whenever you update a code try to come with various different possibilities the changed code might affect the app process. In this case, the issue was clearly visible in the less than 50 lines method code that this PR change belongs to.

The draft I put together for testing is still accessible from draft list with title `Draft for image meta testing`.

https://user-images.githubusercontent.com/6298342/192991235-7c5f5985-18a5-49ed-8dc0-098a70efd49b.mov

src/utils/editor.ts
noumantahir: @aliseyalvi video reference. https://github.com/ecency/ecency-mobile/pull/2473#issuecomment-1261999538

other than this, why not use slice(), instead of using filter(), apparently slice method is less expensive compared to filter.

aliseyalvi:
@noumantahir thumb selection glitch fixed 

https://user-images.githubusercontent.com/48380998/193459802-9c902574-6b30-42c2-adc1-b1a4ea5fb729.mp4

noumantahir (approved):
Now we are talking... :D :D

feruzm:
@noumantahir, @aliseyalvi are we doing same for users and links fields as well?

feruzm (approved):
Lgtm now

PR #2475, opened 2022-09-18
Fix useNativeDriver warnings

### LINKED ISSUE #2464: Refractor: Fix all animated useNativeDriver warnings
better to replace Animated with reanimated package.
with reanimated add easeinout property

aliseyalvi:
### What does this PR?
This PR is WIP for removing/fixing useNativeDriver warnings to either set useNativeDriver to true or replace with React Native Reanimated component
 
### Issue number
fixes #2464


aliseyalvi:
@noumantahir tried to find the useNativeDriver warnings, let me know if I had missed any or find any other warning relared to useNativeDriver

src/components/collapsibleCard/view/collapsibleCardView.js
```diff
@@ -50,6 +50,7 @@ class CollapsibleCardView extends PureComponent {
     Animated.timing(this.anime.height, {
       toValue: this.anime.expanded ? this._getMinValue() : this._getMaxValue() + (moreHeight || 0),
       duration: 200,
+      useNativeDriver: false,
```
noumantahir, line 53: using native driver as false means js thread will be handling the animation. can you please try replacing Animated package with `react-native-reanimated` as we discussed. Start with this one and better replace the package in all places.

noumantahir (changes requested):
The priority is to replace all `import {Animated} from react-native ` with `import Animated, { Easing, Extrapolate } from 'react-native-reanimated';`

If for some reason some implementation of Animated become complicated to migrate to reanimated package, only then set `useNativeDriver` to true. And we should only allow `useNativeDriver` as false if we are out of options completely.

src/components/collapsibleCard/view/collapsibleCardView.js
aliseyalvi: @noumantahir height property is not supported animation on native thread. Tried to find a solution, one solution which I found is to replace height with scaling because scaling supports native animation. I can give that a try again

aliseyalvi: replaced with react-native-reanimated

aliseyalvi:
@noumantahir replaced occurences of Animated from react-native with Animated from react-native-reanimated. ProgressiveImage component is left with native Animated as it is not being reused.

noumantahir (changes requested):
Two issue detected... 

1. App crashes when entering pin code.
Screenshot 2022-09-29 at 2 31 22 PM

2. toast notification opacity is never reaching full opacity, while it is working fine in current development.
Screenshot 2022-09-29 at 2 39 36 PM

Both of these animations can even be changed to use `react-native-reanimatable` instead of reanimated package for simplified and clean code implementation. do whatever seems like a goto solution

aliseyalvi:
@noumantahir replaced reanimated in pinAnimatedInput with animatable and toastNotification as well

src/components/pinAnimatedInput/views/pinAnimatedInputView.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable react/no-array-index-key */
 import React, { Component } from 'react';
 import { View } from 'react-native';
-import Animated, { Easing } from 'react-native-reanimated';
+import * as Animatable from 'react-native-animatable';
```
noumantahir, line 4: better use `import { View as AnimatableView } from 'react-native-animatable'` import * .... is almost always very expensive import

src/components/toastNotification/view/toastNotificaitonView.js
```diff
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import { TouchableOpacity, Text } from 'react-native';
-import Animated, { Easing } from 'react-native-reanimated';
+import * as Animatable from 'react-native-animatable';
```
noumantahir, line 3: better use import { View as AnimatableView } from 'react-native-animatable' import * .... is almost always very expensive import

aliseyalvi: Ok

noumantahir (changes requested):
- toastNotification is working great now. do check one change request I placed
- revert pinAnimatedInput to previous behaviour (before this branch), it was a very refined animation compared to simple fading

aliseyalvi:
@noumantahir reverted changes in pinAnimatedInput

noumantahir (approved):
lgtm

feruzm (approved):

PR #2479, opened 2022-09-20
Nt/use query with drafts

noumantahir:
### What does this PR??
Integrated use query package with drafts api for testing package's caching capabilities, basic api is wrapped around use query framework other than add and update as they being used inside class component, so that part is pending, but added a work around to link and invalidate cached data upon update and addition also left a TODO note to remove workaround later.

For now this can be tested and reviewed before expanding its functions to other api's. Theoretically, this package can solve most of our caching and network related issues we have been facing, but for now it's implementation is kept basic 

- open and play around with drafts
- previously, app would fetch drafts every time we visit drafts page, now it loads once and later it loads in the background and update ui if change is available

### Issue number
https://github.com/orgs/ecency/projects/2#card-85453601

### Screenshots/Video

https://user-images.githubusercontent.com/6298342/191274059-e66e1710-ee02-411b-b9bb-1449e1d6b40e.mov

https://user-images.githubusercontent.com/6298342/191274132-da55289a-982d-489a-baaf-d2b2a7218944.mov

https://user-images.githubusercontent.com/6298342/191274134-6ac833e9-ada4-46c0-a05b-104f253a77ed.mov


feruzm (approved):
lgtm 🚀

PR #2480, opened 2022-09-21
Nt/wallet global props fix

noumantahir:
### What does this PR?|
hot fix, hive balance being returned as zero often times.

Apparently, there is a change user's fetchGlobalProps call either failed on startup or wallet page opened on while fetch was in progress.

This hot-fix makes sure to fetch fresh global props data if not already present before proceeding to aggregating wallet data

This may or may not fix the actually problem but during testing this was the only viable possibility for having 0 values only in HP section while others were populated for user.

### Issue number
https://discord.com/channels/385034494555455488/385448634197737484/1022097808339238955

### Screenshots/Video
**Successful Reproduction**
Screenshot 2022-09-21 at 4 58 48 PM

### DIFF
--- a/src/utils/wallet.ts
+++ b/src/utils/wallet.ts
@@ -1,21 +1,39 @@
 import get from 'lodash/get';
+import { operationOrders } from '@hiveio/dhive/lib/utils';
+import { utils } from '@hiveio/dhive';
 import parseDate from './parseDate';
 import parseToken from './parseToken';
 import { vestsToHp } from './conversions';
-import { fetchGlobalProps, getAccount, getAccountHistory, getConversionRequests, getFeedHistory, getOpenOrders, getSavingsWithdrawFrom } from '../providers/hive/dhive';
+import {
+  fetchGlobalProps,
+  getAccount,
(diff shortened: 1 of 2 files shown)


src/utils/wallet.ts
```diff
@@ -505,36 +492,32 @@ export const fetchCoinsData = async ({
...
   const walletData = {} as any;
 
-
   if (!username) {
     return walletData;
   }
 
+  //fetch latest global props if refresh or data not available
+  const { base, quote, hivePerMVests } =
+    refresh || !globalProps || !globalProps.hivePerMVests ? await fetchGlobalProps() : globalProps;
```
feruzm, line 513: we can cache these globalProps also for couple minutes so it might lower number of requests. or useQuery for retries and cache...

feruzm (approved):
lgtm

src/utils/wallet.ts
noumantahir: will be using useQuery for some the wallet related issue, but useQuery also have limitations, like it is only meant to be used somewhat directly with a component.

PR #2485, opened 2022-09-25
Fixed re-renders warnings in RenderHTML

### LINKED ISSUE #2465: Improve: avoid RenderHTML rerenders
Post renderer is being abused. need to check which component is causing rerenders
image

aliseyalvi:
### What does this PR?
This PR fixes rerenders warnings in PostHtmlRenderer component.
According to maintainers of react-native-render-html this issue is due to some unstable props and they had given a solution to fix the issue. RenderHtml props can be wrapped in usememo to avoid re-renders and fix these warnings

https://github.com/meliorence/react-native-render-html/issues/511

https://stackoverflow.com/questions/68966120/react-native-render-html-you-seem-to-update-the-x-prop-of-the-y-component-in-s

### Issue number
fixes #2465

### DIFF
--- a/src/components/postHtmlRenderer/postHtmlRenderer.tsx
+++ b/src/components/postHtmlRenderer/postHtmlRenderer.tsx
@@ -1,4 +1,4 @@
-import React, { memo } from 'react';
+import React, { memo, useMemo } from 'react';
 import RenderHTML, { CustomRendererProps, Element, TNode } from 'react-native-render-html';
 import styles from './postHtmlRendererStyles';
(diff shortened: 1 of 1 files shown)


noumantahir (approved):
so far in my testing the core purpose of this is being served.. however in my testing video playback is not working, though it may not be related to this PR, so making it a separate issue

Screenshot 2022-09-26 at 12 58 45 PM

feruzm (approved):
lgtm, let's do some tests

PR #2490, opened 2022-09-29
discarding savedInstanceState

### LINKED ISSUE #2489: BUG! New crash 15 Aug onwards

noumantahir:
### What does this PR?
This should apparently fixed abnormal android app crash. It took some good amount of time to pin point the issue, though I have not yet able to reproduce the issue locally. But I am quite certain it is fixed now.

### Issue number
fixed #2489


noumantahir:
And no, we will not be reverting this change, it is in accordance to upgraded react-native-screens documentation

feruzm:
@noumantahir interesting, I was assuming as variable suggest, it is saving state when app goes to background and comes back, maybe it was wrong way of thinking about it then. Good to know, let's see what testing shows 👍

feruzm:
Also does Bugsnag gave some pointers on this?

PR #2491, opened 2022-10-01
fetch activity count when app state is changed from background to active

aliseyalvi:
### What does this PR?
Fetch activity count when app state is changed.
Previously it wasn't getting activity count on app state change and it only happens by manuall refresh or push notification

### Issue number
https://discord.com/channels/@me/920267778190086205/1025419528655601734


noumantahir (approved):
lgtm

feruzm (approved):
Lgtm

PR #2493, opened 2022-10-05
Nt/use query notifications

noumantahir:
### What does this PR?
That took waaay longer then expected, apparently the infinitQuery was way too expensive for us to implement and it was getting extra complicated to control.

However, after three varying implementation attempts, ended up implemented my own version of infinite query, use provided hoo useQueries, so far it was working great for our purpose. right now it is now general purpose and is only being used for notifications, but can easily be modified for general purpose and be used in rest of app, so this initial effort will pay off there.

### Screenshots/Video
In demo video, mark as unread is not working at end, that issue is fixed in latest commits.

https://user-images.githubusercontent.com/6298342/194070579-1d7f94e8-bcba-4f32-b00c-7a201a00f337.mov


feruzm (approved):
lgtm 🚀

PR #2494, opened 2022-10-05
Nt/use query snippets

noumantahir:
### What does this PR?
Integrate use query with snippets, also test tried optimistic updates with snippets update and new addition.

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/194095919-f49acc66-a328-4af9-a938-b60525697b73.mov


feruzm (approved):
lgtm

PR #2498, opened 2022-10-06
[WIP] redefine window dimensions calculation

### LINKED ISSUE #2449: BUG: android app installed on chromebooks not rendering properly
Happens when opening app using an external post link.

Screenshot 2022-09-07 at 7 50 26 PM

Issue might be due landscape/portraint width not calculating properly. Need to reproduce it first

**Update:**
Attempted to reproduce this issue on android table emulator, apparently it is not being reproduce even if we open Ecency app while tablet is locked in landscape mode.

Screenshot 2022-09-23 at 5 19 12 PM

If we notice closely look at the first screenshot shared by melinda, the **BasicHeader** dimensions are being read properly but it messes up as it renders **PostBodyView**. Based on this deduction, the issue seems to be coming from this code linked below, it is just a guess though
https://github.com/ecency/ecency-mobile/blob/development/src/utils/getWindowDimensions.ts#L18

aliseyalvi:
### What does this PR?
This PR changes the way of calculating window dimensions and orientation locking mechanism.
Previously we were calculating window dimensions in a utility method "getWindowDimensions" and inside our theme file. Problem with this approach was on certain devices like android tablets when orientation is set to landcsape and ececny is opened it miscalculate the dimensions because due to our orientation locking mechanism it lcoks the orientation to portrait but dimensions calculated at start doesn't change causing distorted and broken layout.

In this PR, we are using useWindowDimension hook in our main application container and storing the dimensions in redux. And after that using these dimensions from redux into stateful components. I this way whenever orientation changes, useWindowDimensions get aware of dimension changes and it updates redux and redux updates the UI according to new dimensions

### Issue number
fixes #2449 

### Screenshots/Video
Screen record after changes. These changes are only done in header/tabbar and feedscreen.
https://user-images.githubusercontent.com/48380998/194387128-29e27b14-3e33-47f6-9bf4-02eee6f210a0.mp4


aliseyalvi:
@noumantahir this is the video before. When ecency is launched in landscape locked, UI got messed up. And this is because is is taking the landscape dimensions but UI is locked in portrait and UI is not updating because dimensions are not inside stateful component and hence not able to update UI if dimensions get updated

https://user-images.githubusercontent.com/48380998/194388548-be4846c4-811f-4e09-a25e-4f70f5f29dc5.mp4

feruzm:
@aliseyalvi this has some conflicts

aliseyalvi:
@feruzm don't merge this. Maybe we can discard this. This was created to fix renering issue on chromebook and android tablet but that is fixed in separate [PR](https://github.com/ecency/ecency-mobile/pull/2501) with some other trick.

PR #2501, opened 2022-10-09
Bug Fix: app not rendering properly on android tablets 

### LINKED ISSUE #2449: BUG: android app installed on chromebooks not rendering properly

aliseyalvi:
### What does this PR?
In this PR, device info is got at app startup and if device is android tablet then it is locked in the orientation in which app was launched. This fixes the issue of inappropriate layout on android tablets due to msicalculation of dimensions due to change in layout/orientation after app startup. It was the qucikest and easy fix for now to handle issue which was affecting some devices 

### Issue number
fixes #2449


aliseyalvi:
###Before
https://user-images.githubusercontent.com/48380998/194771409-85d63240-363f-4eb5-b2e6-b8aa1e9db409.mp4

aliseyalvi:
### After
https://user-images.githubusercontent.com/48380998/194771508-7cce4f5f-898e-4de5-a8c7-e6af4589fb5d.mp4

aliseyalvi:
> > ### After
> > Monosnap.screencast.2022-10-09.22-42-22.mp4
> 
> seems the trick we discussed actually worked...:D Have you tried launching in portrait mode as just to verify that part is also working as expected.

yes prtrait is also working and this trick is only applied to android tablets, all rest of devices would follow previous method

noumantahir (approved):
this should serve us well temporarily uptill we do not fully support dynamic orientation change.

feruzm (approved):
lgtm

PR #2503, opened 2022-10-11
Nt/long press reply

noumantahir:
### What does this PR?
support for long press to show quick reply without scroll

### Screenshots/Video
https://user-images.githubusercontent.com/6298342/195091875-b9ce4937-19f3-4816-8d5d-2a79d42284a2.mov


feruzm (approved):
Lgtm

PR #2505, opened 2022-10-12
Nt/text input simplify

noumantahir:
### What does this PR?
This pr attempts to simplify text input used in markdown editor. based on Melinda's description of crash it's possible that dynamically changing text input styling might be affecting the stability of app. 

I found some references leading to it but not very promising though, one referenced issue was happening because of fontFamily and on iOS.
ref: https://github.com/facebook/react-native/issues/18815

The overall behaviour of text input is same on both android and ios, also on android oreo specifically.


feruzm (approved):
Let’s try this

PR #2506, opened 2022-10-12
fixed component update from inside function body issue

noumantahir:
### What does this PR?
issue was causing back to back state updates during post details rendering process causing a slight performance drawback.

### Issue number
https://github.com/orgs/ecency/projects/2#card-86165422

### Video
no more  `Cannot update a component from inside the function body of a different component` issue
https://user-images.githubusercontent.com/6298342/195287017-1804bfca-8d3f-4bac-a3f2-435e69c55d94.mov


feruzm (approved):
Good find

PR #2508, opened 2022-10-12
Nt/draft mutation

noumantahir:
### What does this PR?
using react-query to update drafts, optimistic update and save cache. essentially re adds ability to save draft while using is editing post
very enthusiastic live draft update, tested a few times but need well testing.

### Screenshots/Video
No visual change added


feruzm (approved):
lgtm, let's test

PR #2511, opened 2022-10-14
Nt/usr activity cache

### LINKED ISSUE #2510: Cache user activity failed calls
Track failed usr-activity calls and retry them later, no need show usr if it fails, points can be awarded passively.

1. Check for useQuery implantation if we can use it for making failed calls upon app relaunch
2. If not able to use useQuery, go for manual option like cache or redux to call passive activities calls

Update: this thread suggest we might not be able to get what we want from useQuery out of the box and might still have to use some state management
https://github.com/TanStack/query/discussions/3580#discussioncomment-2708403

noumantahir:
### What does this PR?
Keeps a cache of failed user activities calls and calls them on start up and on app resume

### Issue number
fixes #2510

### Screenshots/Video
calls made on startup for failed activities
Screenshot 2022-10-14 at 1 06 58 PM


feruzm (approved):
lgtm

PR #2532, opened 2022-11-15
doing an id based union of notifications while compiling list data

noumantahir:
### What does this PR?
notifications are now filtered based on notification id making sure no repeating id is rendered. 

If there still are duplicated notifications, I would suggest perhaps we should look into the possibility of backend sending duplicated notifications with different ids.

### Issue number
https://github.com/orgs/ecency/projects/2#card-86774211


feruzm (approved):
lgtm

PR #2536, opened 2022-11-17
Nt/bugsnag fixes

### LINKED ISSUE #2535: Alert.alert with non string argument causing app to crash.
Apparently there is some Alert.alert instance in app that is rarely causing app to crash in production, the issue is likely coming because a non string argument is passed to the alert.alert.

bugsnag ref: https://app.bugsnag.com/ecency/ecency-mobile/errors/61096df576a5b40007cffeea?filters[event.unhandled]=true&filters[event.since]=30d&filters[app.release_stage]=production

stackoverflow ref: https://stackoverflow.com/questions/52992606/value-for-title-can-not-be-cast-from-readablenativemap-to-string

noumantahir:
### What does this PR?
1. fixes startup warning caused by bugnsnag apiKey
2. fixes reblogs null crash reported on bugsnag
3. fixes object fed to alert crash reported onbugsnag

Good Thing: there are a lot of bugsnag reports that should settle because our RN upgrade

### Issue number
fixes #2535

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,9 @@
     "bump-patch": "npm version patch --no-git-tag-version",
     "bump-minor": "npm version minor --no-git-tag-version",
     "bump-major": "npm version major --no-git-tag-version",
-    "postinstall": "npx patch-package && node_modules/.bin/rn-nodeify --install --hack --yarn && cd ios && pod install"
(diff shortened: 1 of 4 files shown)


feruzm (approved):
lgtm

PR #2543, opened 2022-11-25
Nt/flipper

noumantahir:
### What does this PR?
After enabling flipper iOS was facing build issues. That is now fixed in this PR.
Also firebase latest package is blocking use of hermes and flipper. Downgraded that to stable 14.12.0.

### Screenshots/Video
Screenshot 2022-11-25 at 4 45 09 PM


package.json
```diff
@@ -165,7 +165,7 @@
     "rn-fetch-blob": "^0.12.0",
     "rn-placeholder": "^1.3.2",
     "rtl-detect": "^1.0.4",
-    "speakingurl": "^14.0.1",
+    "speakingurl": "^1.0.1",
```
feruzm, line 168: accidental change?

noumantahir: seems like it

feruzm (approved):
lgtm

PR #2562, opened 2022-12-30
Nt/drafts cache fixes

noumantahir:
### What does this PR?
After assessing loopholes that might have caused comments to crash app I was able to force reproduce a scenario that Melinda faced i.e. crashing comments got fixed after publishing a previously saved draft.

This is a blind code that might resolve the issue, I will be further assessing the making changes to the drafts cache flow but this can be used as a hotfix


feruzm (approved):
Lgtm

PR #2574, opened 2023-01-09
Nt/prefill staking user

noumantahir:
### What does this PR?
pre-filing destination for staking engine token, but is changeable if user so chose to. 

### Issue number
https://github.com/orgs/ecency/projects/2#card-87513260

### Screenshots/Video
it's a blind code because engine api route is down, but should work fine


feruzm (approved):
Lgtm

PR #2589, opened 2023-01-17
[WIP] Nt/post display optimise

noumantahir:
### Screenshots/Video

initial outcome, all comments and post rendered as single list without nested lists of even scrollviews
https://user-images.githubusercontent.com/6298342/212919564-ebdc0b0e-8091-44e1-a47b-a282a4edbec9.mov


feruzm (approved):
Let’s see, I do hope no navigation got broken, app links, internal etc. will have to test few cases

PR #2594, opened 2023-01-21
Nt/editor hot patch

noumantahir:
### What does this PR?
Initially I started with converting class component to functional in editor, but that complicated things way too much, this is a more soft approach towards patching editor. One think I realised while doing these changes is the editor have become really really messsy prone to leaks very easily. We will have to rewrite editor a clean format no matter.

For now, 

- Removed usage of UNSAFE method by utilising getDerivedStateFromProps
- Avoid react element cloning in editor by bypassing <PostForm /> component and rewiring it's methods
- Loading unsynced draft on priority and also showing it's entry in drafts list.

### Issue number
https://github.com/orgs/ecency/projects/2#card-87699886

### Screenshots/Video
Screenshot 2023-01-22 at 12 28 55 AM


feruzm (approved):
lgtm

PR #2604, opened 2023-01-24
Nt/iap updates

noumantahir:
### What does this PR?
Makes sure user do not accidentally consume account purchase without consent. If unconsumed, user is presented with claim button instead of buy so user can retry a failed purchase either then and there or later at some point, it should even sync between devices through IAP


feruzm (approved):
lgtm

PR #2605, opened 2023-01-24
Nt/boost fixes

noumantahir:
### What does this PR?
I was not able recreate the issue faced by user, but the first step I took ideally fix the issue as this is the only differentiating factor I could find that could lead to an issue if all keys are preset as expected.

- comparing user using username instead of name, as username is separator in selection dropdown as well
- using selected account local data from otherAccounts instead of realm
- improved error messaging toast
- added bugsang to a few relevant places in dhive.js


feruzm (approved):
lgtm

PR #2608, opened 2023-01-25
Nt/non intrusive failure

noumantahir:
### What does this PR?
- using toast messages instead of alert for vote, comment, post failure
- updated bugsnag report for token injection failure


feruzm (approved):
lgtm

PR #2612, opened 2023-01-28
Nt/fixes

noumantahir:
### What does this PR?
- fixed typo issue not letting auto consume enable
- added heic to jpg conversion before upload, require testing on real device.


feruzm (approved):
Lgtm

PR #2617, opened 2023-02-04
remove <span> tags from post content

noumantahir:
### What does this PR?
apparently renderHtml was not treating <span> well if <a> is used inside the and was causing a glitch. since we do not exclusively handle <span> in our mobile side rendering, removing <span> fixed the issue without causing any render differences.

let me know what you think.

### Screenshots/Video
Screenshot 2023-02-04 at 10 24 11 PM


feruzm (approved):
Lgtm, but this might break many other things. Need proper testing

PR #2622, opened 2023-02-23
Nt/comment load fix

noumantahir:
### What does this PR?
fixes discussion fetch failure issue on post screen navigate.

Bug was not happening on every navigate, on first post details navigate if post content was not yet loaded comments would fails to fetch, subsequent navigate was showing comments being loaded as post content was already in cache. permlink was being used from post content fetch response for loading comments.


src/containers/inAppPurchaseContainer.tsx
```diff
@@ -107,7 +107,7 @@ class InAppPurchaseContainer extends Component {
         };
 
         // make sure item is not consumed if email and username not preset for 999accounts
-        if ((purchase.productId === '999accounts') && (!email || !username)) {
+        if (purchase.productId === '999accounts' && (!email || !username)) {
```
feruzm, line 110: Linter issue again? @noumantahir

noumantahir: no, this one should be fine.... this should translate to `( boolean1 && ( boolean2 || boolean3 ))` and that is okay

feruzm (approved):
lgtm

PR #2633, opened 2023-03-07
Nt/processing image ratios

noumantahir:
### What does this PR?
- update image_ratios during new post submit and post edit
- using thumbnail ratio in post feed
- using image ratios in post html rendering
- updated images sorting with custom thumbUrl

### Issue number
https://github.com/orgs/ecency/projects/2#card-88315502

### Screenshots/Video
**BEFORE**

https://user-images.githubusercontent.com/6298342/223436505-f8111f58-2951-4ef9-a50c-778f482484cd.mov

**AFTER**

https://user-images.githubusercontent.com/6298342/223436538-ab13c8e7-0b7e-4a0e-b01e-19cdace97569.mov

### DIFF
--- a/src/utils/editor.ts
+++ b/src/utils/editor.ts
@@ -90,8 +90,8 @@ export const generateReplyPermlink = (toAuthor) => {
   const timeFormat = `${t.getFullYear().toString()}${(t.getMonth() + 1).toString()}${t
     .getDate()
     .toString()}t${t.getHours().toString()}${t.getMinutes().toString()}${t
-      .getSeconds()
-      .toString()}${t.getMilliseconds().toString()}z`;
+    .getSeconds()
+    .toString()}${t.getMilliseconds().toString()}z`;
 
   return `re-${toAuthor.replace(/\./g, '')}-${timeFormat}`;
 };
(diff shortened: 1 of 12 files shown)


src/utils/editor.ts
```diff
@@ -251,6 +253,18 @@ export const extractMetadata = (body: string, thumbUrl?: string) => {
     out.users = matchedUsers.slice(0, 10); // return only first 10 users
   }
 
+  if (out.image && fetchRatios) {
+    out.image_ratios = await Promise.all(out.image.slice(0, 5).map((url) => {
+      return new Promise((resolve) => {
+        Image.getSize(url, (width, height) => {
+          resolve(width / height)
+        }, () => resolve(0))
```
feruzm, line 261: @noumantahir If fail, `resolve(1)` perhaps instead of `0`? Also 2nd slice not necessary?

noumantahir: @feruzm it's intentional, defaulting to 1 would imply a square image while 0 would give a meaning it's ratio is undefined after front-end will fallback to old way to setting ratio.... but if we set it to undefined then array itself would be messed with and might interfere with cross matching `image` and `image_ratios` arrays.

Suggest otherwise if you find this approach troublesome.

noumantahir:
@feruzm PR updated with links and users omitted from metadata....

#fixes https://github.com/orgs/ecency/projects/2#card-87999895

Screenshot 2023-03-10 at 10 43 36 PM

feruzm (approved):
lgtm

PR #2640, opened 2023-03-12
bumped rn version to 0.70.7

noumantahir:
### What does this PR?
RN version 0.70.7 seems to include a fix for TextInputs, though it's not specific if it would fix our issue because the description states it's related to samsung devices, but at this point any change that involve TextInput fixes, will buy it...


feruzm (approved):
Lgtm

PR #2641, opened 2023-03-12
Nt/img cache optimise

noumantahir:
### What does this PR?
- clears image memory on memoryWarning generated, will handle warnings made if feed cache goes beyond limits
- clears image memory coming back from post screen, handles instant memory clearing when post screen unmount, apparently this does not affected images already displayed on screen and re-rendering of posts picks image from disk cache.


feruzm (approved):
lgtm

PR #2647, opened 2023-03-13
exclusive migration token data based on old object existance

noumantahir:
### What does this PR?
If tokens data with old objection formation already preset, app will form new array using the old object instead on local app selection.

This also fixes the issue of having unwanted token making it's way to tokens array in profile data.

### Screenshots/Video
Fabricated test case includes having unavailable token in selection and have non matching selection in profile.tokens and in app selection.
Screenshot 2023-03-14 at 1 20 17 AM
Screenshot 2023-03-14 at 1 21 20 AM

App prioritised tokens object over local selection / other account selected tokens
Screenshot 2023-03-14 at 1 31 58 AM
Screenshot 2023-03-14 at 1 32 18 AM


feruzm (approved):
lgtm

PR #2650, opened 2023-03-14
refined slider value to keep min 2 decimal and max 4 demical digits

noumantahir:
### What does this PR?
if value is greater than 1, 2 decimal digits are fixed
if less then 1, digits will keep under 4 digits and reduce based on value with a precision of 2

### Screenshots/Video

https://user-images.githubusercontent.com/6298342/224962257-7ed1a5c0-8ba3-47bb-bd07-ea42550cb03e.mov

https://user-images.githubusercontent.com/6298342/224962295-61204870-32b2-469b-8de7-9ba6238e389b.mov


feruzm (approved):
lgtm

PR #2689, opened 2023-05-23
Nt/account switch fix

noumantahir:
### What does this PR?
After testing a number of scenarios and trying to recreate the environment being faced by users, added a bunch snippets to help avoid embarrassing situations 

1 - purges malformed accounts
2 - logs out of app instead of logging into malformed account
3 - rearranged logout structure to avoid creating an undefined auth data state
4 - moved isLogingOut state to uiReducer so it gets reset on every app launch if routine is interupted for some reason

### Screenshots/Video
showing toast when auto detecting malformed user data detected on launch and when switching to logged in user after logging out
Screenshot 2023-05-24 at 5 51 40 PM

showing logout prompt if invalid account detected regular account switch
Screenshot 2023-05-24 at 5 54 35 PM

refine text button account switch sheet
Screenshot 2023-05-24 at 6 07 23 PM

using in app action modal instead of OptionsModal 
Screenshot 2023-05-24 at 6 07 31 PM

### DIFF
--- a/src/components/accountsBottomSheet/container/accountsBottomSheetContainer.tsx
+++ b/src/components/accountsBottomSheet/container/accountsBottomSheetContainer.tsx
@@ -92,6 +92,10 @@ const AccountsBottomSheetContainer = ( ) => {
       const realmData = await getUserDataWithUsername(accountData.username);
 
       _currentAccount.username = _currentAccount.name;
+
+      if(!realmData[0]){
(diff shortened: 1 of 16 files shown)


noumantahir:
Added account verification mechanism on startup, account switch and logout if malformed keys data is detected.

feruzm (approved):

PR #2694, opened 2023-05-31
Nt/account repair

noumantahir:
### What does this PR?
This PR attempts to silently fix account data if redux copy and steem data is intact, in some cases either of data might not be preset, for that I have checks in place that will trigger a account verification prompt as was used in previous PR.

There is not visual change to app, but I have tested the flow with force data corruption of users while using trying out both hive signer, master and active keys.

NOTE: the code is tested in simulated account corrupted scenarios, hopefully it should work in real corrupted data as well, but cannot verify till I receive some reports from testers


feruzm (approved):
Lgtm

PR #2695, opened 2023-05-31
Nt - Pin code to keys

noumantahir:
### What does this PR?
if pin code is not already set, app asks user to set pin code before accessing authorities screen. once set, app will ask user to enter pin on subsequent access.

### Screenshots/Video
https://github.com/ecency/ecency-mobile/assets/6298342/50f888d8-da1a-4f1e-ad37-5f64d928a550


feruzm (approved):
Lgtm

PR #2707, opened 2023-06-16
Nt/other accounts repair

noumantahir:
### What does this PR?
PR attempts to repair corrupted other accounts data, keeping the reporter code there to help us track progress


feruzm (approved):
lgtm, let's do some tests

PR #2709, opened 2023-06-17
Nt/sc2 login restructure

noumantahir:
### What does this PR?
While digging into SC2 method, I was able to reproduce the missing data account condition out users might be facing, made some changes in an attempt to both repair and gracefully handle failure.

But keeping in view our recent experience, we should test with multiple logins and logouts with account and switching etc before going for release.

### DIFF
--- a/src/providers/hive/auth.js
+++ b/src/providers/hive/auth.js
@@ -22,6 +22,7 @@ import { getSCAccessToken, getUnreadNotificationCount } from '../ecency/ecency';
 // Constants
 import AUTH_TYPE from '../../constants/authType';
 import { makeHsCode } from '../../utils/hive-signer-helper';
+import bugsnapInstance from '../../config/bugsnag';
 
 export const login = async (username, password) => {
   let loginFlag = false;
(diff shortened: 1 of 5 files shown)


src/providers/hive/auth.js
```diff
@@ -125,13 +126,18 @@ export const login = async (username, password) => {
...
-  const account = await getUser(scAccount.account.name);
-  let avatar = '';
 
-  return new Promise(async (resolve, reject) => {
+  try {
+    const scTokens = await getSCAccessToken(code);
+    hsApi.setAccessToken(get(scTokens, 'access_token', ''));
+    const scAccount = await hsApi.me();
+    
+    //NOTE: even though sAccount.account has account data but there are certain properties missing from hsApi variant, for instance account.username
```
feruzm, line 135: @noumantahir `account.username` is same as `account.name`

noumantahir: essentially username is not always confirm to be present, that is one loophole I detected when I was able to reproduce the missing local issue. name property however is coming from hive itself, so it;s always certain to be there.

feruzm (approved):
lgtm

PR #2710, opened 2023-06-26
Nt/login repair patch

noumantahir:
### What does this PR?
- Updates auth data even if user is already logged in, it would get user out of login roadblock, also allows user to change key type if needed
- rearranged account repair code to avoid early verify prompt
- on successful account repair, continue login routine instead of disrupting it
- during logout, if authData is not present, avoid throwing error, condition is acceptable


feruzm (approved):
lgtm

PR #2714, opened 2023-07-10
Installer package name

### LINKED ISSUE #2682: Detect installation type PlayStore or direct APK
If possible detect installation type PlayStore or direct APK install. If direct APK `new update available` message should direct user to github instead of PlayStore.

moazzambukhari:
### What does this PR?
it can check that in which platform app will be installed
it will be check on testing mode
### Issue number
fixes #2682

### DIFF
--- a/src/screens/application/container/applicationContainer.tsx
+++ b/src/screens/application/container/applicationContainer.tsx
@@ -15,7 +15,6 @@ import changeNavigationBarColor from 'react-native-navigation-bar-color';
 import { connect } from 'react-redux';
 import { injectIntl } from 'react-intl';
 import { bindActionCreators } from 'redux';
-// eslint-disable-next-line prettier/prettier
 import messaging from '@react-native-firebase/messaging';
 import VersionNumber from 'react-native-version-number';
 import ReceiveSharingIntent from 'react-native-receive-sharing-intent';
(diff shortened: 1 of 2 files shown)


src/screens/application/container/applicationContainer.tsx
```diff
@@ -247,13 +248,29 @@ class ApplicationContainer extends Component {
...
-                setLastUpdateCheck(null);
-                Linking.openURL(
-                  Platform.select({
-                    ios: 'itms-apps://itunes.apple.com/us/app/apple-store/id1451896376?mt=8',
-                    android: 'market://details?id=app.esteem.mobile.android',
-                  }),
-                );
+                DeviceInfo.getInstallerPackageName().then((installerPackageName) => {
+                  switch (installerPackageName) {
+                    case 'google-store':
```
noumantahir, line 253: installerPackageName values used in switch statement will always fail... please revisit `getInstallerPackageName` documentation here....

https://www.npmjs.com/package/react-native-device-info#getinstallerpackagename


For example... in case of android play store, the installer package name would be `com.android.vending` and for apple app store it is `AppStore`

noumantahir (changes requested):
package installer is not set up correctly....

noumantahir (approved):
lgtm

feruzm:
> lgtm, @feruzm please verify non store installation redirect link.... `https://github.com/ecency/ecency-mobile/releases`

Yes, that is correct. Of course if we could detect latest release and depending on platform, download it for user, that would be better experience. But let's keep it simple for now.

PR #2723, opened 2023-07-20
Mb/api migration

moazzambukhari:
### What does this PR?
End points that need to migration

- [x] ecency.ts
- [ ] ePoint.ts

### Issue number
#2723


src/providers/ecency/ecency.ts
```diff
@@ -223,9 +224,18 @@ export const deleteBookmark = async (bookmarkId: string) => {
...
+* 
+* body: 
+* type:string
+* data:string
+* 
+**/
 export const addReport = async (type: 'content' | 'user', data: string) => {
   try {
-    const response = await api.post('/report', {
+    const response = await api.post('/private-api/report', {
```
noumantahir, line 238: use `ecencyApi` instead `api`

src/providers/ecency/ecency.ts
```diff
@@ -237,10 +247,21 @@ export const addReport = async (type: 'content' | 'user', data: string) => {
...
+/**
+* TOOD:
+* POST /private-api/request-delete
+* 
+* body: 
+* username:string
+* data:string
+* 
+**/
+export const deleteAccount = async (data: string, username: string) => {
```
noumantahir, line 260: 1. username should be first argument to avoid conflict from previous calls
2. use `ecencyApi` instead `api`

noumantahir (approved):
can be merged for now, reverted api endpoint that are not working properly when migration to private-api..

feruzm (approved):

PR #2752, opened 2023-09-14
Wave duplication fix

noumantahir:
### What does this PR?
I'll be honest, I have no explanation as to why it was happening, compared every possible property that could potentially lead to such duplication, but refining the key extractor with author somehow fixed the issue, tested multiple times with and without fix.

### Issue number?
This Wave repeated 8 times?
https://ecency.com/waves/@ecency.waves/waves-2023-09-13#@hive-179017/re-ecencywaves-2023913t215325736z

### Screenshots/Video
Screenshot 2023-09-14 at 12 10 33 PM
Screenshot 2023-09-14 at 12 11 37 PM


feruzm (approved):
lgtm

PR #2754, opened 2023-09-15
hiding muted waves in real time

noumantahir:
### What does this PR?
hided muted authors support for waves

### Recording
https://github.com/ecency/ecency-mobile/assets/6298342/b8dded26-42b9-4402-a67d-036e11fd6f75


feruzm (approved):
lgtm, next we need to work on community mutes and also downvoted content should be grayed out by default

PR #2755, opened 2023-09-17
[WIP] Improve hive uri QR support

### LINKED ISSUE #2756: Patch resolve transaction, integrate json to identify empty signer field.
Patch resolve transaction, integrate json to identify empty signer field.

aliseyalvi:
### What does this PR?
- added operations.json to validate From/signer field in operation object
- added validation for amount field in operation object
- added error handler to format blockchain exceptions to parse into readable format
- added validation for key required for transaction to perform

### Issue number
fixes #2756 
https://discord.com/channels/@me/920267778190086205/1149266143979245628

### Screenshots/Video

https://github.com/ecency/ecency-mobile/assets/48380998/16b2147c-b3d1-4f0a-9aa8-3a62ce1ffe90

https://github.com/ecency/ecency-mobile/assets/48380998/2bb9bc39-e530-41a8-9933-d42d22a733b8

### DIFF
--- a/src/components/qrModal/qrModalView.tsx
+++ b/src/components/qrModal/qrModalView.tsx
@@ -156,16 +156,26 @@ export const QRModal = ({}: QRModalProps) => {
     }
 
     const parsed = hiveuri.decode(uri);
-    const parsedHiveUriValidation = validateParsedHiveUri(parsed);
+    const authoritiesMap = new Map();
+    authoritiesMap.set('active', currentAccount?.local?.activeKey ? true : false);
+    authoritiesMap.set('posting', currentAccount?.local?.postingKey ? true : false);
+    authoritiesMap.set('owner', currentAccount?.local?.ownerKey ? true : false);
+    authoritiesMap.set('memo', currentAccount?.local?.memoKey ? true : false);
+
+    const parsedHiveUriValidation = validateParsedHiveUri(parsed, authoritiesMap);
     if (parsedHiveUriValidation.error) {
       // show alert to user if parsed uri contains invalid operation data
       Alert.alert(
-        intl.formatMessage({
-          id: parsedHiveUriValidation.key1,
-        }),
-        intl.formatMessage({
-          id: parsedHiveUriValidation.key2,
-        }),
+        intl.formatMessage(
+          { id: parsedHiveUriValidation.key1 },
+          { key: parsedHiveUriValidation.keyType },
+        ),
+        intl.formatMessage(
+          {
+            id: parsedHiveUriValidation.key2,
+          },
+          { key: parsedHiveUriValidation.keyType },
+        ),
       );
       return;
     }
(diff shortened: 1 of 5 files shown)


src/utils/hive-uri.ts
```diff
@@ -1,4 +1,146 @@
...
+  }
+};
+
+// validate and format amount field in operation to 3 decimal places
+const _formatAmount = (amount: string) => {
+  const splitAmt = amount.split(' ');
+  if (
+    splitAmt.length === 2 &&
+    parseFloat(splitAmt[0]) &&
+    (splitAmt[1] === 'HIVE' || splitAmt[1] === 'HBD')
```
noumantahir, line 55: Here we can use our `parseAsset` helper method and we potentially can allow other tokens as well for instance engine token etc....

src/utils/hive-uri.ts
```diff
@@ -1,4 +1,146 @@
...
+  }
+};
+
+/**
+ * Validates parsed data from hive-uri, checks if operation length is not greater than one and __signer is present in operation
+ * Accepts parsed uri from decode method of hive-uri
+ * Returns an object with error status, keys for showing errors, and operation name parsed from operations.json
+ *
+ * */
+export const validateParsedHiveUri = (parsedUri: any, authoritiesMap: Map<string, boolean>) => {
```
noumantahir, line 69: I like the overall approach, I think this method should be named differently maybe, considering it not only validating to uri, it is also formatting it... something like `convertToOpsArray` or 'getFormattedOpsArray' ... and it should only return opsArray... and throw error as promise to keep things more clean... thoughts..?

src/components/qrModal/qrModalView.tsx
```diff
@@ -192,13 +203,19 @@ export const QRModal = ({}: QRModalProps) => {
               id: 'qr.approve',
             }),
             onPress: () => {
-              handleHiveUriOperation(currentAccount, pinCode, uri)
+              handleHiveUriOperation(currentAccount, pinCode, tx)
```
noumantahir, line 206: now that we have started to convert uri to operations array, can we make use of existing op array broadcaster instead of introducing `handleHiveUriOperation`... thoughts...?

noumantahir (changes requested):
direct is good, a few refactoring with testing more and I think we should be good...

src/utils/hive-uri.ts
aliseyalvi: I think hivesigner currently allow only HIVE and HBD, we can add support for other tokens if hivesigner support them

src/components/qrModal/qrModalView.tsx
aliseyalvi: I think handling array operations would add complexity, also hivesigner do not generate uris for array ops

src/utils/hive-uri.ts
noumantahir: parseAsset would still have been handy, that's okay for now...

noumantahir:
@feruzm seems good from my side... please review...

noumantahir (approved):

PR #2765, opened 2023-09-27
Nt/latest wave check

noumantahir:
### What does this PR?
Checks for latest waves on app resume, works in similar way just like posts feed latest post check works.

NOTE: Accidently pushed to development, reverted commits and ported to this new branch, feature needs more testing because revert can tend make things break sometimes, did do it carefully this time though.

### Screenshots/Video
did take screenshot but lost it during revert hassle, :/


feruzm (approved):
Lgtm

PR #2772, opened 2023-10-08
fix: GU-2766 add key prop in flash list

gamingumar:
### What does this PR?

add key extractor in flash list to resolve duplicate items issue.


feruzm (approved):
Lgtm

PR #2774, opened 2023-10-10
separated ImageNotifi, no longer need updating build settings

noumantahir:
### What does this PR?
- removing version auto adapting to primary target version of messaging
- no longer needing build setting updates
- building successfully


feruzm (approved):
lgtm

PR #2776, opened 2023-10-12
avoid rendering reply title

noumantahir:
### What does this PR?
Rendering title only for primary post, avoid rendering title for reply comments and waves

### Screenshots/Video
Screenshot 2023-10-12 at 12 03 36 PM
Screenshot 2023-10-12 at 12 03 46 PM


feruzm (approved):
lgtm

PR #2779, opened 2023-10-17
checking new waves on tab reactive

noumantahir:
### What does this PR?
Logic was already there, just triggering it on tab becoming reactive, ignore first render as first render routines are different. subsequent tab changes will check for updated waves


feruzm (approved):
lgtm

PR #2784, opened 2023-10-19
Nt/comment list keys

noumantahir:
### What does this PR?

- fixes gaps between comments when filter is changed
- disabled animation for first comments layer, apparently recycler view (aka flashlist) is not very fond of delayed animations in list. they tend to repeat causing a very glitchy outcome

### Screenshots
Screenshot 2023-10-19 at 16 31 07


feruzm (approved):
lgtm

PR #2790, opened 2023-11-13
feat: GU-2745 add repeat transfer support for other tokens

gamingumar:


src/screens/assetDetails/screen/assetDetailsScreen.tsx
```diff
@@ -104,6 +104,25 @@ const AssetDetailsScreen = ({ navigation, route }: AssetDetailsScreenProps) => {
     }
   };
 
+  const _handleCoinRepeat = (coinActivity: CoinActivity) => {
+    let parsedUsername = '';
+
+    // from @user1 to @user2
+    parsedUsername = coinActivity.details?.split(' ').pop() || '';
```
noumantahir, line 111: perhaps instead of post processing details string we can extend activity type to support to and from properties, as details string can be subject to change. what do you think...?

noumantahir (changes requested):
Transfer working good for HBD and Engine token... one code suggestion in comments and some issues I detected are...

- [ ]  Ecency Points transfer not working as expected, leads to boost screen in my testing. screenshot 1 and 2
- [ ]  Repeat icon appearing on all activity items, that is only required for transfer operation specifically. screenshot 3
- [ ] HP need to repeat at all, since it's nothing but staked HIVE. screenshot 4



Screenshot 2023-11-13 at 19 57 16

Screenshot 2023-11-13 at 20 05 49

Screenshot 2023-11-13 at 19 56 32

Screenshot 2023-11-13 at 19 57 43

src/screens/assetDetails/screen/assetDetailsScreen.tsx
gamingumar: @noumantahir  correct me If I am wrong, but we still have to use detail string to populate to and from? Also not all tokes support to and from..

gamingumar: I have added support for sender and receiver, but I think it only works for points.

noumantahir: @gamingumar essentially the object we groom/build out of various different types of operations is custom made and can be modified as per requirement... if you check `groomingTransactionData` in /wallet.ts, this is the method we use to make string that eventually we are processing to get preferred username, which can be inconsistent or unnecessary from we can easily get to and from properties out opData of transfer, similar approach should be useable for engine data as well... let me know what you think....

gamingumar: @noumantahir can you check now

noumantahir:
**UPDATE**
instead of using string based operation to decide which operation support repeating, added privilege list of transactions that would support repeating, later this list can be updated to support further operations like swap repeat etc.

https://github.com/ecency/ecency-mobile/assets/6298342/d5fd0d83-fc25-4268-a87f-0d10c005982c

feruzm (approved):
lgtm

PR #2807, opened 2023-12-19
Nt/notifications ordering

noumantahir:
### What does this PR?
- using `gk` property for time passed calculation (as used in ecency-vision)
- refactored component to functional base
- using memo to save uanessacry processing
- using section list instead of flat list, kept the sections same as before (Recent, Yesterday, This Week and so on)
- fixed logic for `Last Week` calculation (probably the primary reason causing sorting issue)
- Renamed Label from `Older than month` to `Last month & older` as older than month seems 30+ days but section starts prior to first of current month.

### Issue number
https://github.com/orgs/ecency/projects/2#card-91379167

### Screenshots/Video
https://github.com/ecency/ecency-mobile/assets/6298342/511e1bf0-8675-4cd8-b7c6-9d22d232545e

### DIFF
--- a/src/components/notification/view/notificationView.tsx
+++ b/src/components/notification/view/notificationView.tsx
@@ -1,33 +1,28 @@
-/* eslint-disable react/jsx-wrap-multilines */
-import React, { PureComponent, useMemo, useRef, useState } from 'react';
-import { connect } from 'react-redux';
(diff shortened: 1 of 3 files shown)


feruzm (approved):
lgtm

PR #2820, opened 2024-01-22
fixed default from value extractor issue, disabled lint checker

noumantahir:
### What does this PR?
the method used for extracting signer filed information apparently started malfunctioning after we did app wide lint fixes. disabled lint for the for loop in method.

There is a way to achieve same result without for loop, but for now it seemed like a simple quick fix

### Issue number
https://discord.com/channels/@me/920267778190086205/1198492766406184971

### Screenshots/Video
Screenshot_2024-01-21_at_10 32 30


feruzm (approved):
lgtm

PR #2822, opened 2024-01-30
Nt/spin game api fixes

noumantahir:
### What does this PR?
- apparently game api calls also needed method arguments update
- also code expects response data in return instead of raw response
- Tested complete game flow as well


feruzm (approved):
lgtm

PR #2825, opened 2024-02-03
add support to process hiveuris from notification data

noumantahir:
### What does this PR?
This code is written in a way spend as little time refactoring code to support hive uri processing directly from notification, we can refactor things once we resume development properly with good funds. I have placed a `TODO` in code to keep us reminded of that

To make it work...

With notification package add the following data...

```
type:"hiveuri"
hiveUri:"hive://sign/op/....."
```
and app should be to decode uri and present appropriate action to be performed

### Screenshots/Video
Screenshot 2024-02-03 235105


feruzm (approved):
lgtm, lets do some tests.

PR #2828, opened 2024-02-06
Nt/annoucements support

noumantahir:
### What does this PR?
- showing first announcement on the list
- path not in use at the moment
- path_link is processed as deeplink by making it full url
- ops is prioritised over path_link
- modal is silenced for 2 days if use close modal or tap Close,
- upon tapping the action button, modal with same id is never displayed to user

### Screenshots/Video
https://github.com/ecency/ecency-mobile/assets/6298342/8d0a5b4e-844c-43d5-8e84-0f21c9924a68


feruzm (approved):
lgtm

PR #2836, opened 2024-02-14
Nt/announcement fix

noumantahir:
### What does this PR?
announcement api call data fix. apparently on android it was working without this step, but this is was the only thing stopped announcement response to be returned. 

Well tested on iOS sim, real device testing will reveal real results

### Screenshots/Video
https://github.com/ecency/ecency-mobile/assets/6298342/e110488f-052c-46d1-bed8-b23659de61dc


feruzm (approved):
lgtm

PR #2842, opened 2024-03-04
patch sb web image

noumantahir:
### What does this PR?
if this fixes the crash issues then that is fine, else ignore for now, thing should ideally get fixed after expo-image, I have high hopes from that...


feruzm (approved):
lgtm

PR #2843, opened 2024-03-04
Nt/rn slight downgrade

noumantahir:
### What does this PR?
based on my findings from TestFlight report and web, apparently some RN upgrades have caused similar issues, downgraded now to 0.73.2  from 0.73.4

Tested this variant on a real iPhone device as well, let's if it survives the production build.


feruzm (approved):
Lgtm

PR #2848, opened 2024-03-06
Nt/expo

noumantahir:
### What does this PR?
basic expo integration

at this stage it does not directly impact anything in app but makes the app capable of supporting expo managed packages moving forward.

purpose of this PR is to start testing app for any instabilities that expo might have caused as my first attempt of integration few days back disturbed the redux persistence for some reason, had to revert all and do again every integration manually.

Initially testing suggests app is stable, but we should build and let testers use app to check stability. meanwhile will integration expo-image, the primary package for which we initially started all the up gradation process.


feruzm (approved):
lgtm

PR #2857, opened 2024-04-16
[Improve] profile edit screen

aliseyalvi:
### What does this PR?

- added avatar update option in edit profile screen
- fixed bug while updating cover image

### Issue number
https://discord.com/channels/@me/920267778190086205/1229321662663495720

### Screenshots/Video

https://github.com/ecency/ecency-mobile/assets/48380998/a9d58860-7a51-4412-8ec3-2a18c4d48d94


src/components/avatarHeader/avatarHeaderView.js
```diff
@@ -1,63 +1,48 @@
...
 import { IconButton } from '../iconButton';
 
 // Styles
 import styles from './avatarHeaderStyles';
 
-const AvatarHeader =
-  ({ username, name, reputation, avatarUrl, showImageUploadActions, isUploading }) =>
-  () => {
-    const navigation = useNavigation();
+const AvatarHeader = ({
```
noumantahir, line 10: if `AvatarHeader` is not being used anywhere else perhaps we can rename component to `EditAvatar`

src/components/avatarHeader/avatarHeaderView.js
```diff
@@ -1,63 +1,48 @@
...
+        <TouchableOpacity onPress={showImageUploadActions}>
+          <UserAvatar
+            key={`${avatarUrl}-${username}`}
+            noAction
+            size="xl"
+            username={username}
+            avatarUrl={avatarUrl}
+            isLoading={isUploading}
+          />
+          <IconButton
```
noumantahir, line 30: Button background should primary blue, and perhaps use `pencil` icon instead `plus/add`, can also be slightly bigger may be.

Screenshot 2024-04-16 at 18 09 07

src/components/avatarHeader/avatarHeaderView.js
```diff
@@ -1,63 +1,48 @@
...
-              </View>
-            </View>
-          </View>
-        </SafeAreaView>
-      </LinearGradient>
-    );
-  };
+        <View style={styles.textWrapper}>
+          {!!name && <Text style={styles.name}>{name}</Text>}
+          <Text style={styles.username}>{`@${username} (${reputation})`}</Text>
```
noumantahir, line 42: this does not make sense here... perhaps we can use similar header as used in profile screen without display name and reputation...

src/components/avatarHeader/avatarHeaderView.js
```diff
@@ -1,63 +1,48 @@
...
+  isUploading,
+}) => {
+  return (
+    <View style={styles.headerContainer}>
+      <View style={styles.wrapper}>
+        <TouchableOpacity onPress={showImageUploadActions}>
+          <UserAvatar
+            key={`${avatarUrl}-${username}`}
+            noAction
+            size="xl"
```
noumantahir, line 25: try using even larger size is possible without making alot of changes

noumantahir (changes requested):
edit functionality working great... left some comment and some feedback below as well...

- After updating cover photo, the cover image still remains locked to older cover in profile details screen. even tried restarting. avatar image instead updated instantly.

- when user updates either profile picture or cover photo, perhaps we should skip the need to tap SAVE button, it should update data automatically, only textual update may require tapping save button after wards.

- can we try enabling cropping for cover picture as well, or perhaps we left it out on purpose for some reason. may be different platform process cover photo differently, need more digging

noumantahir:
@aliseyalvi two things

1. cover image still updating, also tried restarting app and waiting for quite a while
2. after avatar and cover image, profile is updated, that is good... but forcing screen to go back, seems choppy, can make an exception for avatar and cover to not automatically go back.

noumantahir (approved):
lgtm

https://github.com/ecency/ecency-mobile/assets/6298342/fb7033e0-36ff-4dd7-9844-d5c1f83dae45

PR #2861, opened 2024-04-18
Nt/private import fix

noumantahir:
### What does this PR?
After testing login routines for some time I finally identified a loophole that most probably causing login data getting corrupted.

Essentially if someone logged in with active or posting key and attempts to import any of keys under Manage Authorities section. What was happening is it overwrites all of private keys to the one imported by user after login.

This matches perfectly with one of reported by a user facing similar issue of being logged in with active but not able to use it yet not being logged out as well.

This is also directly related to pinCode related issue affecting the login data since pinCode is perquisite of managing authorities in app

### Steps to reproduce
Login with any key other than master password, import some other key under Manage Authorities

### Screenshots/Video
BEFORE
https://github.com/ecency/ecency-mobile/assets/6298342/a134f7f4-1428-4ce8-a2c8-ef4bec7ea322

AFTER
https://github.com/ecency/ecency-mobile/assets/6298342/c5259d66-311d-4f2e-bc4d-4c9752f2e8a6

### DIFF
--- a/src/providers/hive/auth.js
+++ b/src/providers/hive/auth.js
@@ -330,27 +330,27 @@ export const getUpdatedUserData = (userData, data) => {
     masterKey:
       get(userData, 'authType', '') === AUTH_TYPE.MASTER_KEY
         ? encryptKey(data.password, get(data, 'pinCode'))
-        : '',
+        : get(userData, 'masterKey', ''),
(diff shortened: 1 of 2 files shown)


feruzm (approved):
lgtm

PR #2867, opened 2024-04-29
making sure announcement alert appear after welcome modal is closed

noumantahir:
### What does this PR?
Apparently on new update/install, announcement modal was appearing on top of welcome modal, which not only felt annoyed, it might have also caused us to lose possible votes during our last rollout, because at this stage app was not ready to process vote and we might have ended up losing it.

What I also suggest is, after some time of rolling out new update should create a new announcement with different ID and content to encourage users to vote again.

To make it less annoying we can even chose to check if target user already voted, let me know.

### Screenshots/Video
BEFORE
Screenshot 2024-04-29 at 17 26 36

NOW

https://github.com/ecency/ecency-mobile/assets/6298342/857712e7-52fc-44ae-8263-9820a13d4d74

### DIFF
--- a/src/providers/queries/announcementsQueries.ts
+++ b/src/providers/queries/announcementsQueries.ts
@@ -1,14 +1,16 @@
 import { useQuery } from '@tanstack/react-query';
 import { useIntl } from 'react-intl';
 import { useDispatch } from 'react-redux';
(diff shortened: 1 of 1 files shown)


feruzm (approved):
lgtm

PR #2873, opened 2024-05-16
Nt/expo local auth

noumantahir:
### What does this PR?
migrated to expo local auth instead of out dated fingerprint-scanner. working principle of both is same with minor basic exception that are dealt with.

Test on both iOS and Android for now, but require more testing from testers

https://github.com/ecency/ecency-mobile/assets/6298342/99794d79-5d5c-4a87-96fa-cbb0e9c785ec


feruzm (approved):
lgtm

PR #2880, opened 2024-05-27
discarding waves with negative rshare and stats properties

noumantahir:
filters out waves after fetching, right now only applicable one refresh, implementing this live causes a glitch, can be refined though, but for now this is go to solution.

Unable to test with real data as rshares did not change after I downvoted a comment/wave with two of my test users


feruzm (approved):
let's try

PR #2882, opened 2024-05-29
removed conflicting eslint parser

noumantahir:
### What does this PR?
- fixed conflicting lint parser issue
- Sadaqat working fixing pending lint issues

### Screenshots/Video
Screenshot 2024-05-29 at 12 31 42


.eslintrc.json
```diff
@@ -97,6 +97,10 @@
         "ts": "never",
         "tsx": "never"
       }
-    ]
+    ],
+    "class-methods-use-this": "off",
```
noumantahir, line 101: @aliseyalvi do we have to tun these off, no workaround/fix possible...?

aliseyalvi: @noumantahir actually we have to this for every method which is defined in class based component, we have to make them static, so thats why I turned it off because there aren't gonna be class components anymore

aliseyalvi: and for jsx-one-expression-per-line and react/jsx-wrap-multilines, they are actually conflicting with prettier, I mean if I fix them and add parenthesis or move them to next line, when prettier runs it changes them back, so I think its better to turn them off

feruzm (approved):
lgtm

PR #2888, opened 2024-06-28
Crash Bug Fixes  

aliseyalvi:
### What does this PR?
This PR fixes some of the crash bug sources

-  fixed a crash found on android 11 emulator when back camera device is not available 
- fixes a crash source produces when a waves post is deleted just after publishing, it was happening due to query cache not being invalidated

### Issue number
https://discord.com/channels/@me/920267778190086205/1248672460417532037


src/providers/queries/postQueries/wavesQueries.ts
```diff
@@ -355,3 +355,29 @@ export const fetchLatestWavesContainer = async (host) => {
...
+      return wavePermlink;
+    }
+
+    throw new Error('invalid mutations data');
+  };
+
+  const _options: UseMutationOptions<string, unknown, string, void> = {
+    onMutate: async (filteredWavesData: any) => {
+      await queryClient.cancelQueries(QUERIES.WAVES.GET);
+      queryClient.setQueryData([QUERIES.WAVES.GET], [...filteredWavesData]);
```
noumantahir, line 373: @aliseyalvi this might cause issues or it might not do anything at all considering the accumulating wave keys and data is done in a relatively complicated way, check this below snippet for instance.

```
  const wavesQueries = useQueries({
    queries: activePermlinks.map((pagePermlink, index) => ({
      queryKey: [QUERIES.WAVES.GET, host, pagePermlink, index], // index at end is used to track query hydration
      queryFn: () => _fetchWaves(pagePermlink),
      initialData: [],
    })),
  });
```
 
A waves query key is a very combination of base key, host for waves, active container permlink and it's index in list of available containers. we can arrange a short meeting on how it all comes together
  
 Also the effects of cancelQueries also need assessing, for instance what would happen if it is invoked while new containers or waves were being fetched as a result of onEndReached call, will it affect user experience. might be a rare instance though.

src/components/comments/container/commentsContainer.tsx
```diff
@@ -221,6 +224,10 @@ const CommentsContainer = ({
         setLComments(filteredComments);
       } else {
         filteredComments = propComments.filter(_applyFilter);
+        console.log('filteredComments : ', filteredComments);
+        if (postType === PostTypes.WAVE) {
+          deleteWaveMutation.mutate(filteredComments);
```
noumantahir, line 229: possibly we can make a callback in wavesScreen and handle wave delete inside instance of useWavesQuery

noumantahir (changes requested):
waves are slightly bit complicated in comparison to regular api calls and blog posts feed, a simple query data update will not be adequate

noumantahir (approved):
can be merged for now, however we will clean up changes made to wavesQueries in another PR.

PR #2897, opened 2024-07-17
Nt/poll fix

noumantahir:
### What does this PR?
- avoid setting wrong contentType if poll is empty
- ignores empty pollDraft entry, keep poll bar indicator disabled in case of empty poll
- removed POLL status of already published posts with empty data


feruzm (approved):
lgtm

PR #2900, opened 2024-07-27
Nt/wallet api

noumantahir:
### What does this PR?
Migrated wallet processor from using old multiple api call method to using single new portfolio endpoint.
this first iteration only attempts to only fetch all required data from backend and still post processes data on frontend, in next next iteration we would migrate that processing to backend as well and return only a pre-backed data set

One thing not done in this iteration, that might also be something not letting us fully harness the speedy backend response, is the claimable data fetching, later we would also move that part to single api call response

### Issue number
https://github.com/orgs/ecency/projects/4/views/1?pane=issue&itemId=69929988

### Screenshots/Video
Screenshot 2024-07-27 at 17 12 45

**BEFORE**
https://github.com/user-attachments/assets/55ccd635-40fe-4909-9e83-cb4835e4fa9e

**AFTER**
https://github.com/user-attachments/assets/f5a43d10-0688-4965-9132-ec9cd153b0da

### DIFF
--- a/src/components/postCard/children/postCardContent.tsx
+++ b/src/components/postCard/children/postCardContent.tsx
@@ -45,13 +45,12 @@ export const PostCardContent = ({
   }, [dim.height]);
 
   // featured text can be used to add more labels in future by just inserting text as array item
-  const _isPollPost = content?.json_metadata?.content_type === ContentType.POLL && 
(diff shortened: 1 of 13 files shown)


feruzm (approved):
Lgtm

PR #2902, opened 2024-07-30
fixed claim fail toastMessage rendering

noumantahir:
https://github.com/user-attachments/assets/dfed7d56-4c09-45ac-9ae1-0edadedbac3f


feruzm (approved):
lgtm

PR #2906, opened 2024-08-03
[Feat] keep track of previously logged out users

### LINKED ISSUE #2904: keep track of previously logged out users
TO SADAQAT

right now, when user logs out or system state gets expired, the username also disappears. 

Make sure upon logot, the logged in username still appear in account switcher, only in this case it take user back to login screen with username already filled int

aliseyalvi:
### What does this PR?
This PR keep track of previously logged out users and show them in account bottom sheet for easy login. 
User is redirected to login screen with prefilled username

### Issue number
fixes #2904

### Screenshots/Video

https://github.com/user-attachments/assets/7c8236c8-5d90-477a-8594-c5fc2c5665fe


src/components/accountsBottomSheet/view/accountsBottomSheetView.js
```diff
@@ -55,7 +80,7 @@ const AccountsBottomSheet = forwardRef(
           onClose={onClose}
         >
           <FlatList
-            data={accounts}
+            data={[...accounts, ...prevLoggedInUsers]}
```
noumantahir, line 83: to keep data flow more organised, why not added a section below list instead of mixing different objects? can even add a section labeled `Logged out accounts`

src/components/accountsBottomSheet/view/accountsBottomSheetView.js
```diff
@@ -29,20 +32,42 @@ const AccountsBottomSheet = forwardRef(
...
-    // _handlePressAccountTile(item)
-    const _renderAccountTile = ({ item }) => (
-      <TouchableOpacity style={styles.accountTile} onPress={() => switchAccount(item)}>
-        <View style={styles.avatarAndNameContainer}>
-          <UserAvatar username={item.username} />
-          <View style={styles.nameContainer}>
-            <Text style={styles.name}>{`@${item.username}`}</Text>
+    const _handlePressAccountTile = (item) => {
+      if (
+        item &&
```
noumantahir, line 37: this huge logic can be simplified using `?` and `!` operators.

src/components/accountsBottomSheet/view/accountsBottomSheetView.js
```diff
@@ -29,20 +32,42 @@ const AccountsBottomSheet = forwardRef(
...
+              <Text style={styles.name}>{`@${item.username}`}</Text>
+            </View>
           </View>
-        </View>
-        {currentAccount.name === item.username && (
-          <Icon iconType="AntDesign" name="checkcircle" style={styles.checkIcon} size={24} />
-        )}
-      </TouchableOpacity>
-    );
+          {currentAccount.name === item.username && (
```
noumantahir, line 65: There should be an options to remove listed user from logged out users... can add `X` towards end of tile to remove item

noumantahir (changes requested):
left some review comments, functionality wise, seems good.

aliseyalvi:
@noumantahir PR updated according to review

- added option to remove user from list
- added logged out users to separate sections


https://github.com/user-attachments/assets/0981c50d-ba9c-412e-9b43-edef007076b2

aliseyalvi:
@noumantahir updated icon
image

src/redux/reducers/applicationReducer.ts
```diff
@@ -72,6 +79,7 @@ interface State {
   hidePostsThumbnails: boolean;
   isTermsAccepted: boolean;
   isBiometricEnabled: boolean;
+  prevLoggedInUsers: PrevLoggedInUsers[];
```
noumantahir, line 82: not sure if this the right place for `preLoggedInUsers`, for one application store has already overflown, and second since this is more related to account, perhaps it should be in `accountReducer`

Screenshot 2024-08-13 at 15 39 30

noumantahir (changes requested):
One corner case that needs handling is, if I logout of every account and try to login again, there is no way to select from logged out list. it directly lands login screen.

My suggestion would be to open account switch modal at-least at these two places. or only open account switcher if there is at-least one logged out user.

Screenshot 2024-08-13 at 15 44 10
Screenshot 2024-08-13 at 15 43 57

aliseyalvi:
@noumantahir PR updated as per review.

- show account switch modal if user is logged out
- move prev logged out users data from application store to account store

noumantahir:
lgtm

noumantahir (approved):

PR #2907, opened 2024-08-06
updated hook dependencies for better polls loading

noumantahir:
### What does this PR?
apparently on first time poll load, there is usually a gap between with the isLoading flag is updated and query data is available that was causing a permanent loading indicator on poll options.


feruzm (approved):
lgtm

PR #2908, opened 2024-08-15
[Imp] hide charts by default and toggle visibility based on user selection 

aliseyalvi:
### What does this PR?
-  removed chart from wallet screen
-  toggle chart visibility in assetDetailsScreen
-  added toggle animation

### Issue number
https://discord.com/channels/@me/920267778190086205/1273221543874265098

### Screenshots/Video

https://github.com/user-attachments/assets/49e7c95a-4bb7-4221-9fb9-a03c60a96bc1

### DIFF
--- a/src/screens/assetDetails/children/children.styles.ts
+++ b/src/screens/assetDetails/children/children.styles.ts
@@ -143,4 +143,11 @@ export default EStyleSheet.create({
     textDecorationLine: 'underline',
     textDecorationColor: '$primaryDarkText',
   },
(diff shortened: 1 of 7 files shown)


noumantahir (approved):
not very fond of the solution as a long term thing, however should be good for testing wallet crash.

feruzm:
> not very fond of the solution as a long term thing, however should be good for testing wallet crash.

@noumantahir yes let’s test if it helps and we can polish it further if it helps

PR #2909, opened 2024-08-19
Nt/user serach update

noumantahir:
### What does this PR?
primary issue was a missing hook dependency when user change from simple search text to @text, it was only looking for text change before.

### Issue
https://discord.com/channels/385034494555455488/385448617084846085/1274942470345588812

### Screenshots/Video
https://github.com/user-attachments/assets/dfaa56b0-b210-416f-8283-43ddec5d1e35


feruzm (approved):
lgtm

PR #2914, opened 2024-08-27
Nt/beneficiary extraction

noumantahir:
### What does this PR?
Based on my assessments, apparently it is not sorting that was causing the issue but having duplicate beneficiary possibly returned in videoMetadata as we do also inject one default beneficiary manually as well.
Screenshot 2024-08-27 at 17 20 06

Moreover my forced duplication test also produced expected error based on my assessment above. message only talks about sorting order and not pointing toward duplication, which is the issue with my test case.
Screenshot 2024-08-27 at 17 12 09

Updated makeOptions method to discard any duplicated beneficiary along with sorting them out.
Also, this PR handle the passing of already extracted beneficiaries to schedule post routing.

### Issue number
https://discord.com/channels/@me/920267778190086205/1277127392997146665

### Screenshots/Video
Successful publishing of same draft that was failing with existing code.
Screenshot 2024-08-27 at 17 13 42

### DIFF
--- a/src/screens/editor/container/editorContainer.tsx
+++ b/src/screens/editor/container/editorContainer.tsx
@@ -748,6 +748,7 @@ class EditorContainer extends Component<EditorContainerProps, any> {
           fields,
           scheduleDate,
           jsonMeta,
+          beneficiaries,
         });
       } else {
         await postContent(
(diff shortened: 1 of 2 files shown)


feruzm (approved):
lgtm

PR #2915, opened 2024-08-29
Nt/feed query

noumantahir:
### What does this PR?
discarded old `tabbedPostsFetch` helper in favour of react-query based feed manager.
There are not visual changes only internal refactoring.

The overall structure is very similar to the way we handle other paginated feeds like waves and notifications, with added support of feed specific operations.


feruzm (approved):
lgtm

PR #2916, opened 2024-09-03
Nt/refine login flow

noumantahir:
### What does this PR?
instead of bypassing token refresh upon failure, that would have required thorough testing, fine tuned refresh routine and error reporting mechanism.

- added three retries for hsTokenRefresh
- updated error reporting mechanism to share more ordered details in bugsnag


feruzm (approved):
lgtm

PR #2921, opened 2024-09-19
Nt/feed filters

noumantahir:
### What does this PR?
- replaced tab view of feed screen with more maintained package `react-native-tab-view`
- added support for scrollable tab view
- dynamically enable/disable scroll based on number of selected filters
- support for selecting more than 3 filters
- added edit tabs button to right of tabs

Challenges
- ~~using base behaviour of tab bar that comes with tab-view package is making it difficult to add action buttons with tab bar as we have been doing before. We can chose to use or previous tabBar but with that we will lose ability to auto reveal hidden tabs as portrayed in video~~
- response to tab touch and change tabs seems slow, it could either be dev mode issue, or need to add better optimisation, right now only drafted a solution for testing.

### Issue number
https://github.com/orgs/ecency/projects/4/views/1?pane=issue&itemId=80470260

### Screenshots/Video
https://github.com/user-attachments/assets/7d192ece-7a7f-48ea-bddf-fee273a75433


feruzm (approved):
lgtm

PR #2923, opened 2024-09-27
Nt/tab view migration

noumantahir:
### What does this PR?
Replaced ScrollableTabView completely in favour of more maintained `react-native-tab-view`

The UI of tab bar is different from what was implemented before, let me know if it's important to retain previous tab bar behaviour.

### Screenshots/Video
https://github.com/user-attachments/assets/49fd197d-8409-4bb2-a1c0-7869b6b00aac


feruzm (approved):
lgtm

PR #2924, opened 2024-10-03
Nt/orientation tweaks

noumantahir:
### What does this PR?
Laying down a base for tablet support, essentially avoid relying on static dimensions like getWindowDimension and deviceWith constants. 

Also fixes a number of orientation related glitches that were appearing in mobile side, details can be checked in the issue link.

After PR is merged, we should seep the app to see if it did not affect any component formation in a bad way, I did encounter some and fixed them right away, I do expect more but there should not be many ideally.

### Issue number
https://github.com/orgs/ecency/projects/4/views/1?pane=issue&itemId=81468842


feruzm (approved):
lgtm

PR #2926, opened 2024-10-10
Nt/bot comments

noumantahir:
### What does this PR?
- Filtering bot comments from user comments and rendering them as avatar icons at start of comments
- Comments are filtered on initial fetch using bots lists manage by react-query
- In case of not Bot Comments, the whole component will be hidden avoiding any weird blank space

The criticism I have for web implementation is that it does not tell users upfront what exactly these avatars are for... current implementation is closest to web but only with different text indicating these avatar corresponds to bot comments.

**Update**
- introduced a screen to render bot comments instead of rendering in a modal/actionsheet, interacting with comments while rendered in a modal/actionsheet have some issues that needs address, added note regarding the issue that needs addressing in commentsModal component, working on this modal and using it not just for bot component but also for view extra replies will enhance the experience of going through discussion threads

### Screenshots/Video

https://github.com/user-attachments/assets/81babd3a-44bb-41df-864d-a3294079c512


feruzm (approved):
lgtm

PR #2927, opened 2024-10-16
Nt/asking for vote

noumantahir:
### What does this PR?
Request to vote Ecency proposal, asking only those mobile app user who have not yet voted. Support for dismissing request forever, or remind in 3 days in case do dismissal

**Limitation**
The overall experience did not turn out to be how I share the mockup that showed message of courtesy after voting, but settled for only "Thank you" message.
It started taking a lot more time when attempted to fine the behaviour but since it actually a part of flashList, it's not giving me luxury to easily fine tine. However, later we can convert this component to support announcements as well and make it more stable by spending more time to it. right now it does seem worth to spend more time on a component that would only appear once to user.

### Screenshots/Video
https://github.com/user-attachments/assets/6e89d8f1-e2ff-4989-9ad7-5cfedc9088dc


feruzm (approved):
lgtm

PR #2940, opened 2024-11-08
Nt/optional content_type

noumantahir:
removed fallback for setting post content_type, made pollmetadata partial extension


feruzm (approved):
lgtm

PR #2951, opened 2025-01-10
Nt/img picker update

noumantahir:
### What does this PR?
Apparently it seems the upto date picker package work even without media images permissions, a thread suggested to remove permission as the readme documentation seems outdated

Test image and video picking along with profile picture cropping on both latest android SDK 34 and SDK 31 (2021), both seems to be working without issues.

Also tested picker on iOS for update verification.


feruzm (approved):
lgtm

PR #2999, opened 2025-07-28
Nt/upgrade

noumantahir:
### What does this PR?

- Upgraded RN to 0.79.5, app seems to be working stable so far
- Now using API 35 for android for app store to allow update post 31st August
- Enabled New Arch on Android
- Fully migrated to swift, no more obj-c

New Arch on iOS seems to have same bug as before

FIXED
~~There is a new warning appearing after upgrade, triggered from react-query, I have created a thread on their repo, let's see if we get any response, however it does not seem to affect functionality
ref: https://github.com/TanStack/query/discussions/9504~~


noumantahir:
converted back to draft, just identified some items that needs addressing

PR #3012, opened 2025-08-14
Remove bugsnag in favor of sentry

feruzm:


noumantahir (approved):
updated .env related items, rest seems good, app seems to be building and working good from first impressions.

PR #3015, opened 2025-08-20
Migration to TS

feruzm:
Let's do full test to make sure nothing broken


noumantahir (approved):
I used multiple sections of app with this PR code, there seems to no apparent issues

PR #3016, opened 2025-08-24
Nt/gif playback fix

noumantahir:
### What does this PR?
- Switched back to legacy arch on android.
- Managing different gif toggling approach for android and ios.
- handling initial gif load with inView ref

patch made to `withIO` works well with legacy arch but does not work in new arch, since android was using new arch, it was posting that `messureLayout` issue.

The issue causing flash-list to not work with withIO is resolved in flash-list version 2.0, however flash-list version 2.0 is only new arch supported and just does not let the app load if new arch is disabled. I tried to port the ref handling done in 2.0 back in v1 but apparently it's completely rewritten from ground up, and since we will be migrating v2 none the less it made little sense to spend too much in this.

My options were either manage two different branch for android and ios with new and legacy archs or temporarily shift back to legacy arch, for simplicity, I opted for legacy arch for now. in new arch however, with flashList v2 withIO work fine without even needing the patch we made. But iOS TextInput is still unstable for us in new Arch

### Screenshots/Video
The video shows how gifs now reset/restarts when moving in and out of screen on both android and ios
https://github.com/user-attachments/assets/02c62cc8-f73a-4160-b253-e89d10e2f65b


feruzm (approved):
lgtm

PR #3022, opened 2025-08-28
Nt/new arch

noumantahir:
### What does this PR?
- enabled new arch on both iOS and android
- fixed iOS issue causing text editor to malfunction with new arch enabled
- upgraded to flashList v2, new arch support
- used flash list exposed item state management for replies show/hide
- fixed replies clipping issue by using simple reveal animation instead of timed animation
- fixed copy modal crash happening in iOS triggered back to back modal opening.
- controlling gif playback in galleryModal based on visibleItems callback (improving this in a separate PR)
- build script update to control deploy using PR labels


src/utils/post.ts
```diff
@@ -1,18 +1,4 @@
-export default [
```
feruzm, line 1: @noumantahir I assume, we are not using it anymore?

PR #3023, opened 2025-09-09
Nt/viewability tracker

noumantahir:
### What does this PR?

This is actually game changer, implemented our own refined light weight version of intersection observer. 

I have tried other solutions including visibleItemsChange, useContext and old school prop passing, but all tend to affect the component tree or were just extremely expensive to inject in already loaded FlashList

_**A huge round of applause for assisted AI shit, credit must be given where it's due_

**Benefits**
- only measures layout changes for gifs
- only updates the gif whose viewability is changing, leaves the rest of component tree unchanged
- discards unmounted gifs

It can be further refined to better performance as well and a corner case that might trigger one in a million times, but so far it's looking pretty promising, and took a good chunk of our time.

For testing and stability purposes, I have only enabled it for GIFs preset in post content, if it stays stable then we expand with comments, waves and replies.

### Screenshots/Video
iOS
https://github.com/user-attachments/assets/6cc9e20a-b579-4fdf-a5b8-86073f1cb3b8

Android
https://github.com/user-attachments/assets/b1534b88-b7be-4be5-ac3a-0059607e7b9f


feruzm (approved):
lgtm

PR #3025, opened 2025-09-12
Nt/optimise feed - iteration 1

noumantahir:
### What does this PR?
Started with testing feed screen with single flatlist for every filter but apparently the result was same as with using separate tabs with separate flatlist, however I did notice some abnormalities that I am working on next

- clearing image memory cache on tab change (not storage)
I expected flatlist and tabs to manage this internally, but apparently they do not and keep hold to the images cache even if the tab is not used for a while. This small change clears the memory cache, but not clear them from temp storage, means only we can more memory without having any hit on the user experience, or atleast it's not obvious.

- no longer managing image ratio externally
previously we have external maps to manage image ratios for posts not have ratios stored in metadata. this was done so to avoid reinitiated state every time flat list recycles the view. now with new useLayoutState hook, the list manages it's own state for items internally even if the view is recycled, and list will have to listen for changes in one less extraData

Other minor changes
- fixed My Comments section not updating on profile change
- fine tuned feed header paddings a bit
- fixe failing ios CI builds

### Screenshots/Video
BEFORE
Screenshot 2025-09-12 at 11 02 33

AFTER
Screenshot 2025-09-12 at 20 52 54

### DIFF
--- a/.github/workflows/build-ios.yml
+++ b/.github/workflows/build-ios.yml
@@ -22,7 +22,7 @@ jobs:
     steps:
       - uses: maxim-lobanov/setup-xcode@v1
         with:
-          xcode-version: latest-stable
+          xcode-version: '16.4'
 
       - name: Show host machine information
         run: |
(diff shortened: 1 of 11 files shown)


feruzm (approved):
lgtm

PR #3039, opened 2025-09-19
Nt/switch improve

noumantahir:
### What does this PR?
- Added redundancy in account switching, side effect is, it feels a bit slow, but no unexpected behaviour nonetheless
- added switching indicator to make it less boring.
- Avoid stale closure with linking event listener, reinitiates callback on every user change

### Screenshots/Video
https://github.com/user-attachments/assets/bbb1de05-9a3f-45dc-b284-d3ba4ce932a1


feruzm (approved):
lgtm

PR #3140, opened 2026-03-11
fixed all three methods for key upgrade

noumantahir:
The primary issues for hive auth sheet not appearing is the sheets malfunction when trigger one after the other, a hack is to added some delay to make sure second is only triggered after first is down.

However a more better solution is to declare second sheet as dependant on other in sheet manager, may next iteration we will do that if required

https://github.com/user-attachments/assets/cb294a3e-c4e0-4b19-b3eb-60ddc01a6468

https://github.com/user-attachments/assets/57c2dba5-ade5-4e80-bab8-4170490c8836

https://github.com/user-attachments/assets/fff1a489-4740-4524-994b-e299ef25e5ac


feruzm (approved):
lgtm

PR #3222, opened 2026-06-04
fix: stability and correctness fixes for 3.5.7

feruzm:
Stability and correctness fixes ahead of the 3.5.7 release. Lint clean, `tsc --noEmit` clean, full unit-test suite passes.

## Changes
- **Crash hardening** — guard several code paths against malformed/empty inputs (analytics path, market-quote parsing, comment sorting).
- **Announcements** — read the correct account field so announcement banners show for logged-in users again.
- **Locale** — persist the first-launch auto-detect guard only after the language is applied, so an interrupted first launch retries instead of sticking on English.
- **Notifications** — fix transfer notifications not opening the wallet.
- **Feed/Waves** — copy posts before parsing so cached query objects are no longer mutated on refetch (matches the existing single-post/comment paths).
- **External links** — a malformed link now shows an error instead of failing silently.
- **Cleanups** — remove sensitive data from a debug log; minor display fix.

## Validation
- `yarn lint` — clean
- `jest --ci` — 318 passed, 1 skipped

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented crashes from malformed comment sources and currency payloads.
  * Fixed post refresh so content updates reliably.
  * Avoided accidental mutations in feeds/waves that affected displays.
  * Ensured Hive URI processing errors are caught and handled.
  * Announcements now respect the correct account identity.
  * Improved navigation guard to ignore empty-string params.

* **Improvements**
  * Downvoted amounts now show a minus sign.
  * Analytics recording is now non-blocking.
  * Locale detection adjusted to avoid locking defaults.
  * Repair logs no longer expose sensitive key data.

### DIFF
--- a/src/components/comments/container/commentsContainer.tsx
+++ b/src/components/comments/container/commentsContainer.tsx
@@ -85,7 +85,10 @@ const CommentsContainer = ({
   // Component Functions
 
   const _sortComments = (sortOrder = 'trending', _comments) => {
-    const sortedComments = _comments || lcomments;
+    const _source = _comments || lcomments;
+    // Guard against non-array inputs (discussion map / undefined) reaching .sort —
+    // was a top Sentry crash ("undefined is not a function").
+    const sortedComments = Array.isArray(_source) ? _source : [];
 
     const absNegative = (a) => a.net_rshares < 0;
 
--- a/src/components/upvotePopover/container/upvotePopover.tsx
+++ b/src/components/upvotePopover/container/upvotePopover.tsx
@@ -515,7 +515,7 @@ const UpvotePopover = forwardRef(({}, ref) => {
   const downVoteIconName = 'downcircleo';
 
   const _percent = `${isDownVoted ? '-' : ''}${(sliderValue * 100).toFixed(0)}%`;
-  const _amount = `$${amount}`;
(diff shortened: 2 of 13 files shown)


feruzm:
Thanks — addressed the review feedback:

- **converters**: now throws on a missing/partial quotes payload instead of returning `null`, so the caller's existing try/catch keeps the prior values rather than dispatching `{...null}`.
- **announcements**: dependency array now uses `currentAccount?.name` (null-safe).
- **applicationContainer**: report uses `account?.name ?? '<unknown>'`.

On the **feed/waves shallow-copy** note: keeping the shallow copy intentionally. `parsePost` reparses `json_metadata` (a string in cache) onto the copy, and the destructive field swaps land on the copy; the only residual nested write is the idempotent string→object parse of `original_entry.json_metadata`. This matches the existing single-post/comment paths in the same module; a per-item deep clone would diverge from that convention and add cost on every refetch for no real safety gain.

feruzm:
Good catch — also guarded the fallback branch: `if (currentAccount?.name && currentAccount.name === account?.name)` so a null slot in `otherAccounts` can't throw and halt notification registration. (Keeping the feed/waves shallow copy as explained above.)

PR #3224, opened 2026-06-04
fix(wallet): normalize transfer amounts to asset precision across all signing paths

feruzm:
Transfer amounts were not consistently normalized to the asset's required on-chain precision before broadcast, which could cause the node to reject the operation (precision mismatch) or send a wrong quantity. This unifies amount normalization across **all three signing paths** (key, HiveAuth, HiveSigner).

## Changes
- **Shared helpers** in `utils/number.ts` — asset-precision lookup, fixed-precision formatting (no scientific notation), token-quantity formatting, and integer milli-unit conversion — with unit tests.
- **All signing paths** now format the amount to the correct precision before building the op. The HiveSigner path previously applied **no** normalization.
- **Input cap** — the amount field now limits decimal entry to the asset's precision.
- **Hive-Engine** token quantities are formatted without scientific notation / forced padding.
- **SPK/LARYNX** milli-unit amounts are rounded to integers (was a raw float multiply that could yield non-integer values).

## Validation
- `yarn lint` — clean
- `jest --ci` — 334 passed (16 new); updated the op-builder test that asserted the old behavior and added asset-precision / engine / milli-unit / no-exponent cases.

## Notes / follow-ups (out of scope here)
- A few non-wallet-send surfaces (in-app browser bridge, QR/URI signing, token swap) format amounts independently and will be aligned separately.
- The SDK's SPK power/lock builders multiply by 1000 without rounding; the app rounds where it owns the conversion, but a couple of SPK token-transfer/power op contracts should be reconciled with the SDK in a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Amount input now enforces asset-specific decimal precision and disables advancing until token precision loads for tokens.

* **Bug Fixes**
  * Improved decimal precision handling and normalization for native assets, vesting, and tokens so submitted amounts match on‑chain requirements.
  * Token quantities are now consistently formatted to avoid precision/rounding issues.

* **Tests**
  * Expanded test coverage for decimal normalization across transfer types.

### DIFF
--- a/src/containers/transferContainer.ts
+++ b/src/containers/transferContainer.ts
@@ -20,7 +20,7 @@ import { getUserDataWithUsername } from '../realm/realm';
 import { getPointsSummary } from '../providers/ecency/ePoint';
 
 // Utils
-import { countDecimals } from '../utils/number';
+import { getAssetPrecision, toFixedNoExp, formatTokenQuantity } from '../utils/number';
 import { fetchTokenBalances } from '../providers/hive-engine/hiveEngine';
 import TransferTypes from '../constants/transferTypes';
 import { fetchSpkMarkets } from '../providers/hive-spk/hiveSpk';
@@ -263,11 +263,15 @@ class TransferContainer extends Component {
       data.executions = +executions;
     }
 
-    if (countDecimals(Number(data.amount)) < 3) {
-      data.amount = Number(data.amount).toFixed(3);
-    }
+    // Normalize to the asset's on-chain precision before building the op: HIVE/HBD/
+    // POINTS need exactly 3 decimals, VESTS 6; Hive-Engine tokens use their own
+    // precision (no scientific notation). Over-precise amounts are otherwise rejected.
+    const amountValue =
+      tokenLayer === TokenLayers.ENGINE
+        ? formatTokenQuantity(data.amount)
+        : toFixedNoExp(data.amount, getAssetPrecision(fundType));
 
-    data.amount = `${data.amount} ${fundType}`;
+    data.amount = `${amountValue} ${fundType}`;
 
     const _onSuccess = () => {
       dispatch(toastNotification(intl.formatMessage({ id: 'alert.successful' })));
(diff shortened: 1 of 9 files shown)


feruzm:
Good catch — addressed in bff2107.

`toFixedNoExp` (and `formatTokenQuantity`) now **truncate toward zero** instead of rounding, so the broadcast amount can never round *up* past the user's balance and it's consistent with the input cap (which slices). Truncation is done float-safely — format with extra precision first, then slice — so binary-float artifacts (e.g. `10.123 * 1000 === 10122.999…`) can't corrupt the result. Added tests for `10.9995 → 10.999` etc.

feruzm:
Both findings were valid — fixed in dfd2c5c:

**1. Truncation rounded up on a run of 9s (P2).** Confirmed: `toFixedNoExp('1.999999999', 3)` returned `2.000`. The `toFixed(precision+4)`-then-slice approach carried. Replaced with a **string-based** truncation that slices the decimal representation directly — `1.999999999 → 1.999`, and it still keeps float-error values exact (`0.3 → 0.300`, not `0.299`). Added regression tests for both.

**2. Engine token precision hard-coded to 8 (P1).** Confirmed valid (pre-existing, not introduced here — the node rejects the over-precise quantity). Now threads the selected token's `precision` (from `tokenBalance.precision`) into the **input cap** and the engine-quantity formatter, so a 0/3-decimal token can no longer enter or broadcast an over-precise quantity. Falls back to 8 only when precision is unavailable.

Lint clean, `tsc` clean, 335 tests pass.

feruzm:
Both addressed in the latest commit:

- **Points HiveSigner amount (Major):** valid — it was the one path still using `Number(amount).toFixed(3)` (rounds). Now reuses `hsNativeAmount` (truncated to POINT precision), consistent with the other signing paths.
- **Engine submit before precision known (Major):** valid edge case — the submit button is now disabled for Engine tokens until `tokenPrecision` has loaded, so an amount can't be broadcast with the fallback 8-decimal precision.

(The greptile `number.ts` rounding note is from an earlier commit — already resolved by the string-based truncation in `dfd2c5c`.)

Lint clean, `tsc` clean, 335 tests pass.

PR #3227, opened 2026-06-05
fix(auth): make PIN migration and redux migrations fail-safe

feruzm:
Hardens two one-time migration paths so a failure degrades gracefully instead of leaving the app in a bad state.

## PIN re-encryption migration (`migrateUserEncryption`)
- **Await** the re-encryption and **abort on failure** instead of marking the migration complete — so it retries on the next unlock rather than leaving the account stuck with keys on the old PIN.
- **Reuse** the re-encrypted record returned in-memory instead of re-reading storage, avoiding a race with the (un-awaited) write.
- Resolve the underlying pin update even when there is no matching user, so an awaiting caller can't hang.
- The unlock flow no longer advances into the requested operation when the migration didn't succeed.

## Redux-persist migrations
- Wrap each migration so a single failing migration is **skipped** instead of rejecting rehydration and dropping **all** persisted state (which would log the user out and lose local drafts).

## Validation
- `yarn lint` — clean
- `jest --ci` — 336 passed (added a failure-isolation test for the migration wrapper)

Reviewed with an adversarial pass over the auth-unlock and rehydration paths.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed PIN verification process that could hang in certain conditions
  * Improved encryption key update reliability with better error handling
  * Enhanced app stability by gracefully handling update failures and preventing data loss

* **Tests**
  * Added test coverage for update failure scenarios

### DIFF
--- a/src/providers/hive/auth.ts
+++ b/src/providers/hive/auth.ts
@@ -493,8 +493,9 @@ export const updatePinCode = (data) =>
                 currentUser = updatedUserData;
               }
             });
-            resolve(currentUser);
           }
+          // Resolve even when there are no users so an awaiting caller cannot hang.
+          resolve(currentUser);
         })
         .catch((err) => {
           reject(err);
--- a/src/screens/pinCode/container/pinCodeContainer.tsx
+++ b/src/screens/pinCode/container/pinCodeContainer.tsx
@@ -266,12 +266,18 @@ class PinCodeContainer extends Component {
 
     // migrate data to default pin if encUnlockPin is not set.
     if (!encUnlockPin) {
-      await MigrationHelpers.migrateUserEncryption(
+      const migrated = await MigrationHelpers.migrateUserEncryption(
         dispatch,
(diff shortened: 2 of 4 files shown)


feruzm:
Both valid — addressed:

- **No-record guard placement (codex/greptile):** the `!migratedLocal` check now runs **immediately after `updatePinCode` resolves, before** any state mutation. So when storage has no matching user, we bail without nulling `currentAccount.local` or switching the app PIN to `DEFAULT_PIN` — a retry starts from clean, un-touched state.
- **Partial-mutation on mid-migration throw (greptile):** the wrapper now snapshots the (JSON-serializable, persisted) state before running each migration and rolls back to that snapshot on a throw — a true skip rather than a partially-applied mutation. Added a v13 test (mutates `wallet`, then throws on missing `cache`) asserting the partial wallet change is discarded.

337 tests pass, lint clean, `tsc` clean.

PR #3228, opened 2026-06-05
fix(editor): make reply/publish resilient to metadata failures

feruzm:
Hardens the editor reply/publish flow against failures while building post metadata.

## Changes
- **`extractMetadata` → `Promise.allSettled`** — a single failed linked-post fetch no longer rejects the whole metadata extraction (previously this could abort a publish/reply and surface a misleading error). Also removes an undefined `urlData` reference in the error path.
- **`_submitReply` guard** — the permlink/metadata computation now runs inside a try/catch that resets the sending flags on failure, so a metadata error or malformed parent post no longer leaves the reply editor permanently stuck.
- **Android background autosave** — draft autosave on backgrounding now also fires on Android (`background`), not just iOS (`inactive`), preventing loss of unsaved content.

## Validation
- `yarn lint` — clean
- `jest --ci` — 337 passed
- Reviewed with an adversarial pass over the reply/publish flow (no use-before-assign, happy-path metadata shape unchanged, no double-save).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved draft auto-save when the app moves from active to background/inactive on iOS and Android.
  * Enhanced reply submission error handling so failed metadata processing won't leave the editor stuck submitting.
  * Metadata extraction now tolerates individual linked-post fetch failures and continues processing remaining links.

### DIFF
--- a/src/screens/editor/container/editorContainer.tsx
+++ b/src/screens/editor/container/editorContainer.tsx
@@ -307,7 +307,9 @@ class EditorContainer extends Component<EditorContainerProps, any> {
   }
 
   _handleAppStateChange = (nextAppState: AppStateStatus) => {
-    if (this._appState.match(/active|forground/) && nextAppState === 'inactive') {
+    // iOS emits 'inactive' when backgrounding; Android emits 'background'. Handle
+    // both so unsaved draft content is not lost when an Android user backgrounds.
+    if (this._appState.match(/active|forground/) && nextAppState.match(/inactive|background/)) {
       this._saveCurrentDraft(this._updatedDraftFields);
     }
     this._appState = nextAppState;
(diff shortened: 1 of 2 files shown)


feruzm:
Both valid — fixed:

- **Regex false-positive double-save (CodeRabbit):** correct — `'inactive'.match(/active/)` is truthy, so on iOS the `inactive → background` transition saved a second time. Switched to exact equality (`this._appState === 'active'` + explicit `inactive`/`background`).
- **Null fields on background-before-edit (codex):** correct — `_saveCurrentDraft` dereferences `fields`, and `_updatedDraftFields` is `null` until the first edit (this latent crash existed on the iOS path too). The autosave now only fires when there are pending edits.

Both handled at the call site in `_handleAppStateChange`. Lint clean, `tsc` clean, 337 tests pass.

PR #3230, opened 2026-06-05
fix(persistence): correct query-persist allowlist for unified SDK keys

feruzm:
The react-query selective-persistence allowlist had drifted from the unified `@ecency/sdk` query keys, causing both over- and under-persistence.

## What was wrong (verified against runtime SDK keys)
| Query | Runtime key | Was | Now |
|---|---|---|---|
| drafts / schedules | `["posts","drafts"/"schedules","infinite",…]` | persisted **all pages** (full bodies) | first page only |
| notifications list | `["notifications",username,…]` | persisted **all pages** | first page only |
| bookmarks / favorites | `["accounts","bookmarks"/"favorites","infinite",…]` | **never persisted** (matched no branch) | first page only |
| account-full / points / feeds | (unchanged) | correct | unchanged |

The old mobile-specific `switch` cases (`'drafts'`, `'schedules'`, `'notifications'`, `'bookmarks'`, `'favourites'`) were **dead code** after the SDK key unification.

## Changes
- Rewrote `shouldDehydrateQuery` to match the SDK **namespace + subtype**, bounding infinite lists (drafts/schedules/notifications) to the first page like feeds, and persisting bookmarks/favorites again.
- Removed the dead legacy `switch` cases.
- Added a `buster: VersionNumber.appVersion` so persisted data whose shape changed across an app/SDK update isn't restored stale.

## Validation
- `yarn lint` / `tsc --noEmit` — clean
- `jest --ci` — 337 passed
- Verified the policy against every runtime key.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed bookmark and favourite synchronization issues where updates wouldn't consistently appear in all parts of the application
  * Improved data cache handling to automatically refresh stored information when the app is updated to a new version

### DIFF
--- a/src/providers/queries/bookmarkQueries.ts
+++ b/src/providers/queries/bookmarkQueries.ts
@@ -84,7 +84,9 @@ export const useAddBookmarkMutation = () => {
     username,
     code,
     () => {
-      queryClient.invalidateQueries({ queryKey: ['bookmarks'] });
+      // SDK keys bookmarks under the 'accounts' namespace; a ['bookmarks'] prefix
+      // does not match ['accounts','bookmarks',…] so the list would stay stale.
+      queryClient.invalidateQueries({ queryKey: ['accounts', 'bookmarks'] });
       dispatch(toastNotification(intl.formatMessage({ id: 'alert.success' })));
     },
     () => {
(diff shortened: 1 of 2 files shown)


feruzm:
Great catch — valid and directly connected to persisting these lists. Fixed in the latest commit.

The bookmark/favorite **reads** use the SDK `accounts` namespace (`['accounts','bookmarks'/'favorites',…]`), but all four mutation success handlers in `bookmarkQueries.ts` invalidated `['bookmarks']`/`['favorites']`, which don't prefix-match — so after add/remove the list stayed stale (now persisted, so it survived restarts too). Changed the invalidations to the correct `['accounts','bookmarks']` / `['accounts','favorites']` prefixes (matches both the base and the infinite read keys). No other stale `['bookmarks']/['favorites']` invalidations remain in the app.

Lint clean, `tsc` clean, 337 tests pass.

PR #3233, opened 2026-06-05
perf(feed): engage PostCard memo + stabilize interaction handler

feruzm:
Feed/list render-performance cleanups. Behavior-preserving.

## Changes
- **PostCard memo actually engages.** The list passed a fresh inline `handleCardInteraction` arrow per render, and `PostCard`'s memo comparator compares that prop — so every card re-rendered on each list re-render. Now the list passes the stable `_handleCardInteraction` directly, and `PostCard` injects its own `content` via a `useCallback` wrapper. To make the chain truly stable, `_showQuickReplyModal` (a dependency of `_handleCardInteraction`) is now `useCallback`'d, so the handler only changes on login-state change.
- **Leaderboard keyExtractor.** Replaced the random-id fallback (which produced a new key — and a full row remount — every render) with a stable index fallback.
- **Removed two debug logs on hot paths** — one on every scroll-position change, one `JSON.stringify` on every viewable-items change.

## Notes
- Verified end-to-end (adversarial pass) that every feed action (USER/OPTIONS/NAVIGATE/REPLY/UPVOTE/PAYOUT_DETAILS/TIP) receives identical arguments — the wrapper's injected `content` equals the previous `item`, and the UPVOTE `onVotingStart` callback chain is intact.
- The round-2 "comments renderItem defeats memo" item was a false positive (`CommentView`'s comparator only compares data props), so it is intentionally not touched.

## Validation
- `yarn lint` / `tsc --noEmit` — clean
- `jest --ci` — 337 passed

## Summary by CodeRabbit

* **Performance Improvements**
  * Optimized list rendering with stable item key generation and callback handling, reducing unnecessary re-renders when scrolling through posts.
  * Removed debug logging for cleaner performance.

* **Bug Fixes**
  * Fixed list key generation fallback for better list stability during updates.

### DIFF
--- a/src/components/leaderboard/view/leaderboardView.tsx
+++ b/src/components/leaderboard/view/leaderboardView.tsx
@@ -87,7 +87,7 @@ class LeaderboardView extends PureComponent {
           <FlatList
             data={users}
             refreshing={refreshing}
-            keyExtractor={(item) => get(item, '_id', Math.random()).toString()}
+            keyExtractor={(item, index) => (get(item, '_id') ?? index).toString()}
             removeClippedSubviews={false}
             ListEmptyComponent={this._renderEmptyView}
             onRefresh={() => fetchLeaderBoard()}
--- a/src/components/leaderboard/view/leaderboardView.tsx
+++ b/src/components/leaderboard/view/leaderboardView.tsx
@@ -87,7 +87,7 @@ class LeaderboardView extends PureComponent {
           <FlatList
             data={users}
             refreshing={refreshing}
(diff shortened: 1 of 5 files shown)


feruzm:
Good catch — applied. Prefixed the fallback so it can't collide with a numeric `_id` that equals a list index:
`keyExtractor={(item, index) => get(item, '_id')?.toString() ?? \`index-${index}\`}`

(Re the memo change: confirmed the handler is account-agnostic — it changes only on login/logout, not account switch, which is correct since the actions it triggers resolve the current account from Redux at open time.)

PR #3234, opened 2026-06-05
chore: trim react-native-vision-camera patch to the real iOS fix

feruzm:
The `react-native-vision-camera+4.7.3.patch` was **2.2 MB / ~32.5k lines**, but **1292 of its 1293 diffs were accidentally-committed Android build artifacts** (`.cxx/Debug/...` CMake/Ninja cache, `.o`/`.bin` object files, build logs) — captured when `patch-package` ran after an Android build.

Those artifacts made the patch fail to apply, so `patch-package` left `node_modules` pristine and the **one legitimate change silently inactive**: an iOS `takePhoto` guard in `CameraSession+Photo.swift` that ensures the capture session is running (and auto-repairs a dead video connection) before capturing.

## Change
Regenerated the patch with **only** the Swift fix. It now applies cleanly (verified `react-native-vision-camera@4.7.3 ✔` and the guard is present in `node_modules`), shrinking the patch from **2.2 MB → ~2.3 KB**.

The activated guard is a no-op when the session is already running, so normal photo capture is unchanged; it only helps the camera-not-ready edge case (which previously failed regardless).


feruzm:
Thanks — valid, fixed in the latest commit.

**`canAddOutput` guard (the correctness gap):** the auto-repair path now checks `captureSession.canAddOutput(newPhotoOutput)` before adding/reassigning, committing the configuration and rejecting with `cameraNotReady` if it can't be added — so `self.photoOutput` can never be left pointing to an output not attached to the session.

**Blocking `usleep` on the camera queue:** leaving as-is intentionally. It only runs on the recovery path (camera already not-ready/degraded), where briefly blocking to recover is preferable to failing outright; making it fully non-blocking is a larger native refactor (KVO on session state) and out of scope for this patch-cleanup. Happy to file a follow-up if you'd prefer.

(Note: CodeRabbit couldn't review this PR — it hit a usage/rate limit, not a code issue.)

PR #3246, opened 2026-06-10
fix(vote-slider): stop thumb slipping to 100% mid-drag on Android

feruzm:
## Problem

Users reported the upvote slider **slipping** — while dragging, the thumb jumps forward and **sticks at 100%** for a moment, even though the finger is still mid-track. Reproduced from a user screen-recording (Android, "show touches" on): the finger indicator stays around the middle of the track while the rendered value snaps to 100% and holds ~0.5s before recovering.

## Root cause

The drag-anywhere slider mapped finger → value with:

```ts
const ratio = evt.nativeEvent.locationX / trackWidth;
```

On Android, `locationX` on PanResponder **move** events intermittently arrives relative to a parent/window origin instead of the track itself. Frame-by-frame measurement of the recording (track left ≈ 122px, width ≈ 216px) matches the symptom exactly: a bad sample adds `trackLeft / trackWidth` (≈ +0.565) to the ratio, which then clamps to `1.0`:

- true 64% → 120% → **clamps 100%**
- true 53% → 109% → **clamps 100%**

The `$estimate` and `NN%` labels are variable-width siblings of the `flex: 1` track, so the wider `"100%"` string shrinks the track — a positive feedback loop that makes the overshoot *stick* until the finger moves far enough left to break out. `pointerEvents="none"` on the fill/thumb does **not** prevent this.

## Fix

- Map finger position from the gesture's **absolute window X** (`gestureState.moveX` / `x0`) minus the track's `measureInWindow()` window-left. `moveX` and `measureInWindow()` share the same window coordinate space, so the ratio is unambiguous regardless of which native view the OS hit-tests under the finger. The track origin is re-measured on layout and at each gesture start.
- Give the `$estimate` and `NN%` labels stable min-widths (`-100%` is the widest percent value) so the flex track no longer resizes mid-drag — removing the residual reflow jitter and keeping the ✎ pencil from jumping around.

Normal-case behaviour is unchanged; only the glitch path is eliminated.

## Testing

- `yarn lint` clean on the changed files; types resolve (`PanResponderGestureState`, `measureInWindow`).
- ⚠️ Needs an **on-device Android** check (the input glitch is Android-specific and doesn't reproduce in a simulator): drag up and down repeatedly across the 50–90% range — the thumb should track the finger with no jumps to 100%. iOS regression check that drag + tap-to-position still set the expected value.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved upvote slider drag responsiveness and accuracy by using absolute gesture coordinates and stable track measurements.
  * Stabilized slider layout during dragging with fixed minimum widths and enforced text alignment.
  * Prevented vote amount and percentage labels from wrapping by constraining them to a single rendered line.

### DIFF
--- a/src/components/upvotePopover/children/upvoteStyles.ts
+++ b/src/components/upvotePopover/children/upvoteStyles.ts
@@ -43,15 +43,25 @@ export default EStyleSheet.create({
     shadowOpacity: 0.35,
     elevation: 3,
   },
+  // minWidth keeps the estimate from reflowing the flex slider track as the
+  // value changes while dragging (the track width/origin feeds the gesture
+  // math in voteSlider.tsx); left-aligned so it still reads naturally.
   amount: {
     fontSize: 10,
     fontWeight: 'bold',
     color: '$primaryDarkGray',
     marginLeft: 8,
+    minWidth: 42,
+    textAlign: 'left',
   },
+  // Fixed-width readout: the percent label is the slider track's right-hand
+  // sibling, so a stable width ("-100%" is the widest value) stops the track
+  // from resizing mid-drag and keeps the ✎ pencil from jittering.
   percent: {
     fontWeight: 'bold',
     color: '$primaryDarkGray',
+    minWidth: 44,
+    textAlign: 'center',
   },
   slider: {
     flex: 1,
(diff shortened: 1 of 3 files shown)


feruzm:
Addressed bot feedback in `5be6540`:

- **Async measurement on gesture start** (CodeRabbit / Greptile P2): `onPanResponderGrant` requested a fresh `measureInWindow()` but mapped `x0` synchronously, so the tap-down used the previously-cached origin. `_measureTrack` now takes an `after` callback (with a synchronous fallback when the native measure is unavailable) and the grant maps `x0` from inside it, so the first emitted value uses up-to-date geometry. Move events were already safe (`widthRef` pre-populated by `onLayout`, guarded against zero).
- **Guard consistency** (Greptile P2): width guard tightened to `w > 0`.

Note: the failing **CodeQL** checks (Analyze ruby/actions) are the repo-wide default-setup scans unrelated to this TS-only diff; **Lint & Unit Tests** passed.

PR #3254, opened 2026-06-13
fix(wallet): unblock Hive Engine token transfers and harden HE reads

feruzm:
## Problem

Hive Engine (HE) token transfers — e.g. ARCHON — could not be sent: the **NEXT button stayed disabled** for every HE token, with a valid recipient, amount, and memo. Some users also saw the wallet intermittently fail to load.

Root cause: the transfer screen gates NEXT on the token's precision being known, but precision was read from the HE **balances** table, which has no `precision` field — it lives on the **tokens** table. So `tokenPrecision` was always `undefined` and the gate never passed. (Even when it didn't block, an unknown precision let an over-precise quantity reach the sidechain, where it is silently rejected.)

Separately, all HE reads go through a single proxy with no request timeout or retry, so a transient blip surfaced as an empty wallet or a stuck transfer — while the same operation works on hive-engine.com (which fails over across nodes).

## Fix

**Transfer (precision):**
- Source Engine precision from the **tokens** table (`fetchTokens`) instead of the balances row; fetch balances + token info in parallel. Precision `0` (integer tokens) is preserved, not defaulted away.
- Thread token precision through the HiveAuth op-builder path (`buildTransferOpsArray` → `getEngineActionOpArray` → `getEngineActionJSON`) so the broadcast quantity is truncated to the token's real precision on **all** auth types (key / HiveSigner / HiveAuth), not the 8-decimal fallback.

**Reliability (wallet load):**
- Add a request timeout + bounded retry-with-backoff to the HE proxy reads.
- Surface wallet-load failures with a **Retry** affordance instead of a blank list, and cap the portfolio query's retry backoff so it can't sit on the skeleton.

## Tests

- New unit tests covering engine precision threading (over-precise truncation, integer/precision-0 tokens, 8-decimal fallback when precision is unknown).
- `yarn lint` clean; `transactionOpsBuilder` / `number` / `wallet` suites green (110 tests).

## Notes

- Pure-JS change — no native rebuild required.
- Follow-up (not in this PR): poll the HE sidechain tx status after broadcast so a rejected transfer stops reporting "Successful".

## Summary by CodeRabbit

* **New Features**
  * Wallet now shows a refresh-failure banner with a "try again" button when data fails to load.

* **Improvements**
  * Network requests use per-request timeouts and sequential retries with a capped retry delay to surface errors faster.
  * Token transfers now preserve and honor explicit token precision (including zero) for more accurate amounts; whole-number engine-token sends are allowed while precision is loading.

* **Tests**
  * Added tests covering engine token precision handling.

### DIFF
--- a/src/containers/transferContainer.ts
+++ b/src/containers/transferContainer.ts
@@ -117,10 +117,16 @@ class TransferContainer extends Component {
         // over-precise (sidechain-rejected) quantity. Fetch both and source
         // precision from the token definition. Precision can legitimately be 0
         // (integer tokens), so keep it as-is rather than defaulting a falsy 0 away.
-        const [tokenBalances, tokens] = await Promise.all([
+        // allSettled so a failure in one leg doesn't discard the other: a token-
+        // metadata (precision) outage shouldn't hide an already-fetched balance, and
+        // a balance outage shouldn't hide precision. Each read rejects (rather than
+        // resolving []) on a real proxy failure, so a rejected leg is left unset.
+        const [balancesResult, tokensResult] = await Promise.allSettled([
           fetchTokenBalances(username),
           fetchTokens([fundType]),
         ]);
+        const tokenBalances = balancesResult.status === 'fulfilled' ? balancesResult.value : [];
+        const tokens = tokensResult.status === 'fulfilled' ? tokensResult.value : [];
 
         enginePrecision = tokens.find((t) => t.symbol === fundType)?.precision;
 
--- a/src/containers/transferContainer.ts
+++ b/src/containers/transferContainer.ts
@@ -21,7 +21,7 @@ import { getPointsSummary } from '../providers/ecency/ePoint';
 
 // Utils
 import { getAssetPrecision, toFixedNoExp, formatTokenQuantity } from '../utils/number';
-import { fetchTokenBalances } from '../providers/hive-engine/hiveEngine';
+import { fetchTokenBalances, fetchTokens } from '../providers/hive-engine/hiveEngine';
 import TransferTypes from '../constants/transferTypes';
 import { fetchSpkMarkets } from '../providers/hive-spk/hiveSpk';
 import TokenLayers from '../constants/tokenLayers';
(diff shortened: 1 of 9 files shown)


feruzm:
Addressed the review feedback in `3f24933`:

- **NEXT gate blocking on unknown precision** (CodeRabbit): the precision gate now only applies to *fractional* Engine amounts. A whole-number amount is precision-safe at any precision, so it's allowed through even if the token-metadata lookup degrades — the common case no longer dead-buttons NEXT, while fractional amounts still wait for precision so an over-precise quantity can't reach the sidechain.
- **Engine reads swallowing failures into `[]`** (CodeRabbit): `fetchTokenBalances`/`fetchTokens` now reject on a real transport failure (after the retries) and resolve `[]` only for a genuine empty result, so the wallet query and `assetsSelect` can retry/surface the error instead of treating an outage as an empty wallet. The transfer balance fetch uses `Promise.allSettled` so one leg failing doesn't discard the other.
- **Dead `fetchHiveEngineTokenBalances`/`fetchMetics`** (CodeRabbit): removed — they had no callers (so they weren't actually aborting any wallet load), along with their now-unused imports.
- **Error UI hidden when stale data is cached** (Greptile): the load-failure Retry is now a banner above the list rather than `ListEmptyComponent`, so a failed *refresh* of an already-loaded wallet is surfaced too, not just the first-load empty case.

Lint/types clean; engine precision threading unit tests still green.

feruzm:
Both follow-up findings were real — fixed in `452774b`:

- **`Number.isInteger(amount)` always false** (CodeRabbit + Greptile): `amount` is string state, so the whole-number degraded-mode escape never activated — engine integer transfers stayed blocked when precision was unknown, defeating the intended fallback. Now `Number.isInteger(Number(amount))`. (The existing `amount > 0` clause still blocks empty/`"0"`, so coercing `""`→`0` is harmless.)
- **`alert.try_again` doesn't exist** (CodeRabbit): `try_again` lives under `ai_assist`, not `alert`, so the label would render a raw id. Switched to the existing `alert.something_wrong_reload`. (Note: the IntlProvider loads only the active locale with no en-US fallback, so a new en-US-only key wasn't an option — reused an existing key present in all 39 locales.)

One deliberate deviation: I kept `result ?? []` on `fetchTokenBalances`/`fetchTokens`. A 200 response with no `result` is a *genuine* empty wallet (new accounts) and must stay `[]`; only transport failures now reject (the blanket `.catch` is gone). Dropping `?? []` would make callers iterate over `undefined`.

PR #3267, opened 2026-06-18
fix(profile): stop the profile header scroll oscillation

feruzm:
## Problem
On a profile page, scrolling the feed makes the page **repeatedly snap back to the fully-expanded top** (avatar/bio/follower-count reappear) ~once a second — the scroll position is unstable ("flakiness"). Reproduced from a user screen-recording; affects **all tabs** (posts/waves/replies/wallet).

## Root cause
`ProfileView._handleOnScroll` flips `isSummaryOpen` via `setState` on every feed scroll event with **no dead zone** (collapse on any `offsetY>0`, expand on `offsetY<=0`). That one flag (a) toggles the top `<Header>` and (b) animates `CollapsibleCard`'s height over 500ms — both **above the `FlashList` in a flex column**. Each toggle reflows the list; FlashList v2 on Fabric momentarily reports `offsetY≈0` during the resize, which **re-fires the opposite toggle** → a self-sustaining ~1s oscillation. `onScrollBeginDrag` also toggled, so even a stationary touch kicked it off.

## Fix (this PR — pure scroll-handler logic, zero layout/structure change)
- **Hysteresis / dead zone:** collapse only when `offsetY > 80`, re-expand only when `offsetY <= 8`. Jitter/overscroll around 0 can no longer flip state.
- **Cooldown:** ignore further flips for 550ms after a toggle (≥ the 500ms `CollapsibleCard` animation), so the reflow that animation causes can't sneak a flip in mid-animation.
- **No begin-drag toggle:** `handleOnScrollBeginDrag={null}` — a touch-start / tiny drag no longer collapses; real movement (`onScrollEndDrag` + per-tab `onScroll`) still does.

This breaks the feedback loop that produced the oscillation, with no render-structure change (trivially revertible).

## Known follow-up (separate PR, needs device QA)
The top `<Header>` is still conditionally **mounted/unmounted** on collapse — a one-frame column-height step that can still cause a *single* jump on a legitimate collapse (not the oscillation). The robust end state is an **always-mounted overlay header driven by a Reanimated shared `scrollY` with a fixed list `paddingTop`**, so the list box never reflows. Scoped and ready to do next.

## Testing (on device — FlashList v2/Fabric scroll behavior is device-sensitive)
- [ ] Profile page: slow-drag the feed past ~80px → header collapses **once** and stays; keep scrolling → no snap-back to the expanded summary.
- [ ] Scroll fully to top → header re-expands **once**; tiny rubber-band at top does **not** cause flicker.
- [ ] Fast flick down/up repeatedly → no ~1s oscillation.
- [ ] Repeat on **every** tab: POSTS, WAVES, REPLIES, and (own profile) WALLET.
- [ ] Switch tabs while scrolled; pinned-post profile; pull-to-refresh on waves/comments — header state stays consistent.
- [ ] `yarn lint` + `yarn test:ci`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profile header scroll behavior to eliminate rapid expand/collapse flickering during header animations. Added explicit scroll thresholds and a cooldown window to stabilize expand/collapse decisions, including consistent behavior after manual summary expansion.

### DIFF
--- a/src/components/profile/profileView.tsx
+++ b/src/components/profile/profileView.tsx
@@ -42,6 +42,10 @@ const SUMMARY_TOGGLE_COOLDOWN_MS = 550;
 class ProfileView extends PureComponent {
   _lastSummaryToggleAt = 0;
 
+  _lastOffsetY = 0;
+
+  _summaryRecheckTimer = null;
+
   constructor(props) {
     super(props);
     this.state = {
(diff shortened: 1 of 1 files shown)


feruzm:
Addressed the review feedback:

- **coderabbit (`handleOnScrollBeginDrag={null}`)** — the prop is typed optional (`?: () => void`), so `null` violated the contract. **Omitted the prop entirely** (begin-drag is intentionally not wired; `onScrollEndDrag` drives collapse).
- **codex P2 — "Recheck top offsets after the cooldown expires"** — fixed. `_handleOnScroll` now records the latest offset and, when a wanted toggle lands inside the 550ms cooldown, schedules a one-shot re-check after the cooldown clears using the most recent offset. So returning to the top no longer gets stuck collapsed even if no further scroll event arrives. Timer is cleared in `componentWillUnmount`.
- **codex P2 — "Forward momentum scrolls before disabling begin-drag"** — acknowledged, **deferred to the redesign follow-up** (not fixed here). Collapse is driven by `onScrollEndDrag`; tracking momentum requires forwarding the continuous `onScroll`, but that path lives in the shared `TabbedPosts`/`postsTabContent`, which is also used by the feed, tag-result and community screens (community has the same collapse pattern). Routing continuous scroll there would change scroll-handler behavior for those screens too, so it's out of scope for this targeted fix. The planned redesign (per-list Reanimated shared `scrollY` with a fixed header inset) tracks momentum natively and removes this gap. Worst case today: a fast flick released before the 80px threshold collapses one gesture late — a minor responsiveness nit, not the reported oscillation.

feruzm:
Good catch on `_handleOnSummaryExpanded` — fixed. The tap-driven expand path now stamps `_lastSummaryToggleAt = Date.now()` before `setState`, consistent with the scroll-driven toggle, so a tap-expand while scrolled past the threshold no longer leaves the cooldown stale (no immediate re-collapse on the next scroll event).

PR #3268, opened 2026-06-18
fix(android): offload FCM message-store write off main thread (Background ANR)

feruzm:
## Problem — the Background-ANR family (ECENCY-MOBILE-9 + 6W/EJ/1D/58/7C)
Confirmed from **live Sentry ANR stacks** (not theory):
```
io.invertase.firebase.messaging.ReactNativeFirebaseMessagingReceiver.onReceive
  -> ReactNativeFirebaseMessagingStoreImpl.storeFirebaseMessage
  -> UniversalFirebasePreferences.getStringValue/setStringValue
  -> SharedPreferencesImpl.getString/edit -> awaitLoadedLocked          (load-side ANR)
... and at onPause/onStop:
android.app.QueuedWork.processPendingWork -> waitToFinish -> SharedPreferencesImpl.writeToFile   (flush-side ANR)
```
react-native-firebase messaging persists its **own** SharedPreferences message store on the **main thread** inside the FCM `BroadcastReceiver` (no `goAsync()`). Every notification does a main-thread prefs read+write; `QueuedWork` then flushes the `.apply()` finishers on the main thread at lifecycle transitions → ANR.

**Not AsyncStorage** — verified `@react-native-async-storage/async-storage@2.2.0` is SQLite-backed (`RKStorage`) on Android and never touches SharedPreferences/`QueuedWork`. (So the deferred AsyncStorage→MMKV migration would *not* have fixed this; it stays a separate, next-update perf initiative.)

## Fix — `patch-package @react-native-firebase/messaging@23.4.0`
In `ReactNativeFirebaseMessagingReceiver.onReceive`:
- Keep `notifications.put(...)` (the in-memory map) **synchronous** on the main thread — it's read first by `getInitialNotification`/`onNewIntent`, so **notification-tap deep-linking is unaffected**.
- Offload **only** the `storeFirebaseMessage` SharedPreferences read/write to a single-thread `Executor` via `goAsync()` (keeps the receiver alive ~10s until `finish()`).

This removes **both** ANR frame families (the `awaitLoadedLocked` load and the `QueuedWork` flush). Android-only, no public-API/JS change, iOS untouched. The patch was generated from the exact v23.4.0 source and verified to apply (`git apply --check`).

Also corrects the misleading comments in `store.ts` / `queries/index.ts` (added in #3260) that attributed the ANR to an AsyncStorage flush — the cache cap + throttle there are SQLite write-volume hygiene, not the ANR fix.

### Why this over the alternatives
- **Firebase upgrade**: no release through 24.1.1 moves `storeFirebaseMessage` off the main thread; 23.8.x pulls Firebase iOS SDK 12.10.0 (the pod/deployment-target friction we already hit). Rejected.
- **eNotify data-only**: would bypass the store but break Android quit-state `getInitialNotification` deep-linking and trade this ANR for a worse per-message cold-boot ANR. Rejected as the fix (payload-trim/burst-rate reduction remain useful zero-risk interim levers, no app release needed).
- **Config toggle**: none exists that disables the store while keeping `getInitialNotification`.

## Testing (on device — required; I could not build/patch-apply here)
- [ ] `yarn` (postinstall runs patch-package): confirm **`react-native-firebase+messaging+23.4.0` applied** with no warning.
- [ ] **Tap deep-link regression (critical)** — foreground, background-tap, and **quit-state cold-start tap** all route correctly via `onNotificationOpenedApp`/`getInitialNotification` (`_pushNavigate` for vote/mention/follow/reblog/reply…).
- [ ] **ANR repro** — low-end Android: burst 20-50 notifications while backgrounding/foregrounding; confirm no ANR and no `awaitLoadedLocked`/`QueuedWork` main-thread stalls (vs an unpatched build).
- [ ] StrictMode disk-write check: `storeFirebaseMessage` no longer logs main-thread disk IO.
- [ ] Watch Sentry ECENCY-MOBILE-9/6W/EJ/1D/58/7C decline post-release; confirm no new cold-boot ANR family appears.
- [ ] iOS smoke (patch is Android-only).

## Follow-ups (not in this PR)
- Optional defense-in-depth: lower `MAX_SIZE_NOTIFICATIONS` (100→~20) in `ReactNativeFirebaseMessagingStoreImpl` to shrink any overgrown prefs file.
- ⚠️ Re-generate + re-QA this patch on **any** `@react-native-firebase/messaging` version bump (patch-package warns if it stops applying).

## Bug Fixes
* Improved notification handling performance by offloading notification persistence work to a background thread, reducing the risk of UI stalls during incoming messages.
* Strengthened notification storage cleanup by lowering the maximum retained notification count and ensuring pruning continues until the stored list is back within the limit, even if it had grown beyond the cap.

### DIFF
--- a/patches/@react-native-firebase+messaging+23.4.0.patch
+++ b/patches/@react-native-firebase+messaging+23.4.0.patch
@@ -1,0 +1,42 @@
+diff --git a/node_modules/@react-native-firebase/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingReceiver.java b/node_modules/@react-native-firebase/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingReceiver.java
+index cee80e8..a6b9178 100644
+--- a/node_modules/@react-native-firebase/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingReceiver.java
++++ b/node_modules/@react-native-firebase/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingReceiver.java
+@@ -16,6 +16,11 @@ public class ReactNativeFirebaseMessagingReceiver extends BroadcastReceiver {
+   private static final String TAG = "RNFirebaseMsgReceiver";
+   static HashMap<String, RemoteMessage> notifications = new HashMap<>();
+ 
++  // Ecency patch: run the message-store SharedPreferences read/write off the main thread to
++  // avoid the awaitLoadedLocked load + QueuedWork.apply() flush Background ANRs (see onReceive).
++  private static final java.util.concurrent.ExecutorService STORE_EXECUTOR =
++      java.util.concurrent.Executors.newSingleThreadExecutor();
++
+   @Override
+   public void onReceive(Context context, Intent intent) {
+     Log.d(TAG, "broadcast received for message");
+@@ -32,9 +37,22 @@ public class ReactNativeFirebaseMessagingReceiver extends BroadcastReceiver {
+     // Add a RemoteMessage if the message contains a notification payload
+     if (remoteMessage.getNotification() != null) {
+       notifications.put(remoteMessage.getMessageId(), remoteMessage);
+-      ReactNativeFirebaseMessagingStoreHelper.getInstance()
+-          .getMessagingStore()
+-          .storeFirebaseMessage(remoteMessage);
++      // Ecency patch: the in-memory notifications.put above stays synchronous (it is read
++      // first by getInitialNotification / onNewIntent). Offload only the SharedPreferences
++      // store read+write off the main thread; goAsync() keeps the receiver process alive
++      // (~10s budget) until finish() is called.
++      final PendingResult storePendingResult = goAsync();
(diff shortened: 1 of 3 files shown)


feruzm:
Addressed both review notes in the patch (single push to avoid re-review churn):

- **Silently-swallowed exceptions (`:38`)** — added `catch (Throwable t) { Log.e(TAG, "Failed to store FCM message off the main thread", t); }` before the `finally`. Off-thread store failures are now logged, and the single-thread executor's worker isn't terminated by an uncaught throw.
- **`goAsync()` 10s deadline under bursts (`:39`)** — mitigated by bounding the store: `MAX_SIZE_NOTIFICATIONS` lowered 100→20 and pruning changed from a single removal to a `while`-loop (so an already-overgrown prefs file shrinks to the cap in one pass). Each store read/write stays small (~ms), so even a large serial burst on the single-thread executor settles far under the 10s `goAsync` window. Kept the executor single-threaded (ordering + no thread churn); the writes are small enough that a bounded/drop-oldest queue would add risk without benefit.

Patch regenerated from the exact v23.4.0 source and re-validated with `git apply --check` (now a 2-file patch: receiver offload + store bounding). Still needs on-device QA per the PR checklist (tap deep-link across fg/bg/quit + burst ANR repro).

PR #3273, opened 2026-06-18
feat(waves): reply-count chip on wave cards (main-feed style)

feruzm:
Follow-up to the waves engagement work. The wave action row had a comment icon with **no count** — the reply count only showed as the "X more replies" expander. Now that esync provides `children`, show a discrete reply-count chip in the action row.

## Change
Replace the count-less comment `IconButton` in `CommentView` with a `TextWithIcon` reply-count chip:
- **Same style as the main feed** comment count (`postCardActionsPanel`): `TextWithIcon` with `comment-outline` + the count.
- Matches the adjacent vote-count chip on the wave card (`iconSize={20}`, `wrapperStyle=leftButton`, `textStyle=voteCountText`) so votes and replies look consistent.
- No longer login-gated — the count is engagement data (like the vote count), shown to everyone. Tapping still opens the reply composer via `_handleOnReplyPress`, which guards login internally.

`text={childCount || 0}` (0 when none), matching the main feed comment chip. Prettier clean.

## New Features
* Reply counts are now displayed on comments, showing the number of responses to each comment. This count is visible to all users regardless of login status.

### DIFF
--- a/src/components/comment/view/commentView.tsx
+++ b/src/components/comment/view/commentView.tsx
@@ -217,16 +217,16 @@ const CommentView = ({
           textStyle={styles.voteCountText}
         />
 
-        {isLoggedIn && (
-          <IconButton
-            size={20}
-            iconStyle={styles.leftIcon}
-            style={styles.leftButton}
-            name="comment-outline"
-            onPress={_handleOnReplyPress}
-            iconType="MaterialCommunityIcons"
-          />
-        )}
+        <TextWithIcon
+          iconName="comment-outline"
+          iconSize={20}
+          wrapperStyle={styles.leftButton}
+          iconType="MaterialCommunityIcons"
+          isClickable
+          onPress={_handleOnReplyPress}
+          text={childCount || 0}
+          textStyle={styles.voteCountText}
+        />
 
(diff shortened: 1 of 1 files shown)


feruzm:
Thanks — both addressed in 24cb37ae9c.

**1. Reply count shown twice (depth-1).** Resolved, but the other way around from the suggestion: hiding the chip on `_depth === 1` would hide it on the **main waves feed** (depth-1 cards) where the expander is the only count surface — which is exactly where this chip was requested. So instead I kept the chip everywhere and **reduced the expander to just its toggle arrow** (dropped its `{childCount} more replies` label), so the number renders once (in the chip) while the expander keeps its expand/collapse function.

**2. Silent no-op for logged-out tap.** Gated the chip on `isClickable={isLoggedIn}`, so for logged-out users it is a **display-only count** (non-interactive, no dead tap) while the count stays visible to everyone. Reply still requires login (the chip composes for logged-in users via `_handleOnReplyPress`); a login prompt on tap would be a larger cross-cutting change since the same silent guard exists on the tip button, so I kept it consistent rather than special-casing this one path.

PR #3286, opened 2026-06-23
fix(a11y): make the upvote control screen-reader accessible

feruzm:
## Investigation

A user who relies on a screen reader reported (via @zumzulik's post) that they **cannot upvote posts** because "the arrow is not read by the software."

Confirmed on the mobile app: the shared `UpvoteButton` (`src/components/postCard/children/upvoteButton.tsx`) renders the vote control as a `TouchableOpacity` wrapping an icon, with **no accessibility metadata** — no `accessibilityRole`, `accessibilityLabel`, `accessibilityState`, or `accessibilityHint`. An icon carries no accessible name, so VoiceOver/TalkBack announce the control as an unlabeled element (or skip it), and the user can't tell what it is or that it's actionable. (App-wide, only ~9 files use `accessibilityLabel`, so a11y is generally sparse — but the upvote control is the one the report is about.)

`UpvoteButton` is shared by **feed cards** (`postCardActionsPanel`), **comments** (`commentView`), and the **single-post action bar** (`postDisplayView` — the bar in the screenshot), so one fix covers all three.

## Fix

On the vote `TouchableOpacity`:
- `accessibilityRole="button"`
- `accessibilityLabel` — state-aware: **Upvote** / **Upvoted** / **Downvoted**
- `accessibilityState={{ selected: isVoted || isDownVoted }}` — so the reader announces the current vote state
- `accessibilityHint` — "Double tap to open vote options" (a tap opens the vote slider/popover)

Also added `accessibilityRole="button"` + label to the payout-details button. New strings under the `post` i18n namespace.

## Also bundled: `@ecency/sdk` 2.3.22 bump

This PR also bumps `@ecency/sdk` `^2.3.21` → `^2.3.22` (`package.json` + `yarn.lock`), pulling in the Waves "For you" feed fix from vision-next #992 (now merged + published). That fix guards the per-container `bridge.get_discussion` call in `getThreads` so a single transient RPC failure skips that container instead of collapsing the whole feed into "Nothing here". Patch bump, no SDK dependency changes; `yarn install --frozen-lockfile` verified. (Carried here because the originally-planned host PR, #3285, already merged.)

## Files

- `src/components/postCard/children/upvoteButton.tsx`
- `src/config/locales/en-US.json` (5 new `post.*` strings)
- `package.json`, `yarn.lock` (SDK bump)

## Validation

- Prettier 2.8.8 (CI-pinned) ✓ · ESLint ✓ (0 errors; 2 pre-existing warnings unrelated to this change) · valid JSON · `yarn install --frozen-lockfile` ✓
- Manual: please confirm on device with VoiceOver (iOS) / TalkBack (Android) — focusing the vote control should announce e.g. "Upvote, button" and "Upvoted, selected, button" after voting.

## Follow-ups (out of scope)

The same treatment would benefit the other action-row controls (reblog, comments, bookmark, gift/tip) and other interactive icons across the app — a broader a11y pass is worth scheduling.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@babel/preset-typescript": "^7.26.0",
     "@babel/runtime": "^7.26.7",
     "@ecency/render-helper": "^2.5.11",
-    "@ecency/sdk": "^2.3.21",
+    "@ecency/sdk": "^2.3.22",
     "@esteemapp/react-native-autocomplete-input": "^4.2.1",
     "@esteemapp/react-native-multi-slider": "^1.1.0",
     "@native-html/iframe-plugin": "^2.6.1",
--- a/src/components/postCard/children/upvoteButton.tsx
+++ b/src/components/postCard/children/upvoteButton.tsx
@@ -92,10 +92,10 @@ export const UpvoteButton = ({
   // Give the vote control a screen-reader name + state. The icon alone carries no
   // accessible label, so VoiceOver/TalkBack users couldn't find or operate it.
   const voteAccessibilityLabel = isVoted
-    ? intl.formatMessage({ id: 'post.upvoted' })
+    ? intl.formatMessage({ id: 'post.upvoted', defaultMessage: 'Upvoted' })
     : isDownVoted
-    ? intl.formatMessage({ id: 'post.downvoted' })
-    : intl.formatMessage({ id: 'post.upvote' });
+    ? intl.formatMessage({ id: 'post.downvoted', defaultMessage: 'Downvoted' })
+    : intl.formatMessage({ id: 'post.upvote', defaultMessage: 'Upvote' });
 
   return (
     <View style={styles.container}>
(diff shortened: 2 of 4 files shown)


feruzm:
Addressed both findings in 0c6d07b:

**Payout amount hidden from screen readers (Codex/Greptile P2):** correct — an explicit `accessibilityLabel` on the payout `TouchableOpacity` overrode the child currency `Text`. Switched it to `accessibilityHint`, so the formatted amount stays the announced name (e.g. "$1.23, button") and "Payout details" is read as the hint.

**Raw message IDs for non-English locales (Codex/Greptile P2):** added `defaultMessage` to every `formatMessage` in the component, so locales that don't yet have the new `post.*` keys fall back to readable English labels instead of strings like "post.upvote". Kept it local rather than reconfiguring `IntlProvider` globally.

(Also note: this PR now bumps `@ecency/sdk` → 2.3.22, pulling in the Waves "For you" fix from vision-next #992.)

PR #3309, opened 2026-06-27
fix: Waves tab icon alignment + draft caret follow-ups

feruzm:
Follow-ups to #3308 (already merged).

## Waves tab icon alignment
The wavy-dash Waves icon sat ~15px higher than the other bottom-tab icons. Cause: `react-native-svg`'s `<Svg>` ignores the shared `paddingTop` that the vector-font icons honor on their `<Text>`. Fix: apply that layout offset to a wrapping `<View>` (which honors padding) and let the parent tab cell center it horizontally, so the icon keeps the same footprint as the others across phones and tablets on iOS and Android.

> Verifying visually on an iOS simulator; will tune by a pixel if the optical center needs it before this merges.

## Draft caret follow-ups (review findings)
- **Missing caret now restores to the end of the body, not position 0.** The body input auto-focuses for returning users, so a `0` default risked the keyboard opening at the top and prepending on the next keystroke. This also makes the caret feature purely additive — legacy drafts, drafts created on another device, and first opens after the feature shipped keep the long-standing end behavior; only drafts with a saved caret change.
- **Cancel the debounced caret persister on unmount.** Prevents a pending 600ms write from firing after publish/first-save has cleared the entry (which would re-create an orphaned `caretMap` entry).

## Testing
- `yarn lint` clean, `tsc` clean (pre-existing tsconfig error only), editor reducer tests pass.
- On-device alignment check across iPhone / iPad / Android still recommended before/after merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved caret restoration in the markdown editor: when no valid saved caret exists, the cursor now restores to the end of the draft instead of jumping to the start.
  * Ensured pending editor updates are committed on close, preventing delayed caret writes that could cause unexpected cursor behavior when reopening drafts.

* **Refactor**
  * Updated icon rendering so layout, spacing, and padding behave consistently by applying styles to a wrapping container.

### DIFF
--- a/src/assets/svgs/wavy-dash-icon.tsx
+++ b/src/assets/svgs/wavy-dash-icon.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { StyleProp, ViewStyle } from 'react-native';
+import { StyleProp, View, ViewStyle } from 'react-native';
 import Svg, { Path } from 'react-native-svg';
 
 interface Props {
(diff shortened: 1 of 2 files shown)


feruzm:
Both review comments point at the same unmount cleanup, and they pull in opposite directions:

- **CodeRabbit (Major):** also handle `_debouncedOnTextChange` on unmount (the body debounce has the same late-fire window).
- **Greptile (P1):** `_persistCaret.cancel()` *drops* a pending caret if the cursor moved within the 600ms window → stale caret on reopen.

Resolved both with `.flush()` instead of `.cancel()` (d37b034):

```js
useEffect(
  () => () => {
    _persistCaret.flush();
    _debouncedOnTextChange.flush();
  },
  [_persistCaret, _debouncedOnTextChange],
);
```

Flushing commits the pending caret **and** body synchronously during the unmount cleanup (which runs before the parent's unmount save), so:
- Greptile: the last cursor move is persisted, not dropped — no stale caret on reopen.
- CodeRabbit: the body debounce is handled too, and runs synchronously rather than firing after unmount; it also stops losing the last ~500ms of typing.

The only residual is a harmless, rare orphan `caretMap` entry if the cursor moved <600ms before publish (publish clears the entry, then the flush re-adds it for an already-published draft) — bounded and never read, which I judged a better trade than Greptile's user-facing stale-caret.

PR #3311, opened 2026-06-27
feat(waves): show tip count + already-tipped state on waves feed cards

feruzm:
## What

Show a wave's **tip count** on the waves feed cards, and mark waves the viewer has **already tipped** — matching the web waves feed.

## Why

Tip counts were only visible on the single-post view. In the waves feed the gift action showed no count, so a tipped wave looked untipped and you couldn't tell whether you'd already tipped it while scrolling.

## How

Backed by esync + `@ecency/sdk` 2.3.26, which add `tip_count` and `tipped_by_viewer` to each waves feed item (`tipped_by_viewer` is computed from the `observer` the feed already sends — the logged-in user).

- Bump `@ecency/sdk` `^2.3.25` → `^2.3.26` (adds the tip fields to the `Entry` type; the values already flow through the feed response and `parsePost` at runtime).
- `commentView`: the gift action becomes a `TextWithIcon` showing `tip_count`, and swaps to a solid `gift` icon (from `gift-outline`) when `tipped_by_viewer` is true — the at-a-glance already-tipped signal, mirroring web.

## Notes

- Reuses the existing `post.a11y_tips` / `post.a11y_tip` strings (no new i18n).
- Backward compatible; the single-post view already shows the count.

## Summary by CodeRabbit

* **New Features**
  * Enhanced the comment action area for tips: when a tip count is available, the UI now displays the current count and a clearer, count-based accessibility label; otherwise it remains icon-only.
  * The tip icon and presentation update based on whether the viewer has already tipped.
* **Bug Fixes**
  * After sending a tip from feed cards, related Waves feed tip counters refresh to reflect the latest state.
* **Chores**
  * Bumped the app’s `@ecency/sdk` dependency to the latest patch version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@babel/preset-typescript": "^7.26.0",
     "@babel/runtime": "^7.26.7",
     "@ecency/render-helper": "^2.5.16",
-    "@ecency/sdk": "^2.3.25",
+    "@ecency/sdk": "^2.3.26",
     "@esteemapp/react-native-autocomplete-input": "^4.2.1",
     "@esteemapp/react-native-multi-slider": "^1.1.0",
     "@native-html/iframe-plugin": "^2.6.1",
--- a/src/components/comment/view/commentView.tsx
+++ b/src/components/comment/view/commentView.tsx
@@ -256,29 +256,47 @@ const CommentView = ({
           })}
         />
 
-        {isLoggedIn && (
-          <TextWithIcon
-            iconName={comment.tipped_by_viewer ? 'gift' : 'gift-outline'}
-            iconSize={20}
-            wrapperStyle={styles.leftButton}
-            iconType="MaterialCommunityIcons"
-            isClickable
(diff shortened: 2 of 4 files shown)


feruzm:
Addressed both findings in 4c9c1b4:

- **Misleading `0` in non-waves views (Greptile P2):** `commentView` is shared with regular comment threads, which don't carry `tip_count`. Now the count + solid-gift state render **only when `tip_count` is a number** (the waves feed); everywhere else the gift falls back to the original icon-only action, so no `0` appears where the value is unknown.
- **Stale count after tipping from a feed card (Greptile P1 / Codex P2):** `useSendTipMutation.onSuccess` now also invalidates the waves feed query (`['posts','waves','feed']`, prefix-matching every variant) alongside the post-tips + wallet invalidations, so a tip sent from a card refreshes its count and already-tipped state instead of waiting for the next feed refetch. The single-post view already updated via the post-tips query.

PR #3353, opened 2026-07-21
Record check-in on content open instead of app startup

feruzm:
Moves the daily check-in (point activity type 10) off app startup, where it competed with everything else the app does on launch, back to the moments that indicate the user is actually reading.

Check-in now fires when:
- a post, comment or reply is opened, including from the notifications screen (all go through the post screen)
- the waves feed is focused

A shared `useCheckIn` hook holds the 15 minute client-side spacing that matches the backend, so repeated opens send at most one request per window per account. The app open and foreground-return calls in `useInitApplication` are removed.

## Summary by CodeRabbit

- **New Features**
  - Added automatic check-ins when opening an existing post and when the Waves screen is focused.
  - Check-ins now require an authenticated account with a username.
  - Repeated check-ins are limited to once every 15 minutes.

- **Bug Fixes**
  - Reduced duplicate check-ins by throttling repeated attempts.
  - If a check-in fails, the throttling window is cleared to allow retries.
  - Continued updating pending activities on app foreground return (without triggering check-ins during app initialization).

### DIFF
--- a/src/components/postView/view/postDisplayView.tsx
+++ b/src/components/postView/view/postDisplayView.tsx
@@ -92,12 +92,15 @@ const PostDisplayView = ({
   );
 
   // Component Life Cycles
+  // Reading a post, comment or reply (incl. opened from notifications) is a check-in.
+  // Not mount-only: a cold start from a deep link can render this before persisted
+  // auth is restored, and `recordCheckIn` changes identity once the account lands,
+  // so the check-in still gets recorded. Its own throttle keeps repeats cheap.
   useEffect(() => {
-    // reading a post, comment or reply (incl. opened from notifications) is a check-in
     if (!isNewPost) {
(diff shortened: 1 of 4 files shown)


feruzm:
Both review points addressed in ff1cda5:

- **Late auth hydration** (greptile P1 / coderabbit): the post-view effect is no longer mount-only. It now depends on `recordCheckIn`, whose identity changes once the persisted account lands, so a cold start from a notification or deep link still records the check-in. The hook's own throttle keeps the extra runs free.
- **Failed attempt consuming the window** (greptile P1): the throttle slot is still claimed before firing (so parallel opens don't double-post), but released in an `onError` if the request fails after retries, so the next post or waves open can try again instead of waiting out the 15 minutes. The queued replay path is unchanged.

PR #3358, opened 2026-07-22
fix(upload): stop sending "undefined" as the image upload signature

feruzm:
Origin logs on the image server show `POST /hs/undefined` from `okhttp/4.12.0`, answered `400 {"error":{"name":"bad_request"}}` — 21 requests across 7 distinct clients over Jul 21-22. Affected accounts fail every upload; unaffected ones are fine (371 successful uploads from the same client on the same day).

`signImage()` can return `undefined` two ways:

- `isHsClientSupported(authType)` is true and `decryptKey()` returns undefined (it is typed `string | undefined` and swallows the failure)
- neither branch matches — no stored `postingKey`, and authType is not STEEM_CONNECT/HIVE_AUTH/ACTIVE_KEY — so the function falls off the end

No caller checked the result, and `imageApi.upload()` interpolated it straight into `baseURL: ${NEW_IMAGE_API}/hs/${signature}`.

## Changes

- `signImage()` throws `SIGN_IMAGE_UNAVAILABLE` instead of returning `undefined`
- `upload()` rejects on a falsy signature, so the placeholder cannot reach the server regardless of caller
- Callers map the error to the existing `alert.decrypt_fail_alert` copy ("Corrupt app state, please relogin to reset state.") rather than a generic failure. `profileEditContainer` also had the `signImage` call outside its `try`/`catch`, so a throw there would have left the upload spinner running; chat no longer reports these through `reportMediaPickerError`.

The retry in `useMediaUploadMutation` no longer re-sends a request that cannot succeed — a thrown error has no `response.status`, so it is not retried.

## Testing

`src/config/imageApi.test.ts` asserts no request is built for an absent signature, and that a present one produces a signature-scoped URL. Verified the guard test fails against the unpatched `upload()` and passes with it. Full suite: 660 passed, 1 skipped. Lint clean on touched files (existing warnings only).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved image upload reliability when signing credentials are unavailable.
  - Prevented uploads from being attempted without a valid image signature.
  - Added clearer, localized alerts for credential or decryption-related failures.
  - Preserved existing handling for canceled image selections while providing more specific errors for signing failures.

### DIFF
--- a/src/components/uploadsGalleryModal/container/uploadsGalleryModal.tsx
+++ b/src/components/uploadsGalleryModal/container/uploadsGalleryModal.tsx
@@ -19,6 +19,7 @@ import { MediaItem } from '../../../providers/ecency/ecency.types';
 import { SpeakUploaderModal } from '../children/speakUploaderModal';
 import { SheetNames } from '../../../navigation/sheets';
 import { selectIsLoggedIn } from '../../../redux/selectors';
+import { isSignImageUnavailable } from '../../../constants/imageUpload';
 
 export interface UploadsGalleryModalRef {
   showModal: () => void;
@@ -396,7 +397,13 @@ export const UploadsGalleryModal = forwardRef(
           const errorMessages = new Set<string>();
           failures.forEach((failure) => {
             const error = failure.status === 'rejected' ? failure.reason : failure;
-            if (error.toString().includes('code 413')) {
+            if (isSignImageUnavailable(error)) {
+              errorMessages.add(
+                intl.formatMessage({
+                  id: 'alert.decrypt_fail_alert',
+                }),
+              );
+            } else if (error.toString().includes('code 413')) {
               errorMessages.add(
                 intl.formatMessage({
                   id: 'alert.payloadTooLarge',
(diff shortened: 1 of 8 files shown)


feruzm:
Addressed all three review findings in 62e6f0df6f — all verified against the code first, and all three were real.

**Codex (P2, editorQueries/uploadsGalleryModal)** — confirmed. `_handleMediaOnSelected` aggregates rejected uploads and falls through to `error.message`, so a sign failure surfaced as a second alert reading literally `SIGN_IMAGE_UNAVAILABLE`. Worth noting this was a regression against the previous behaviour: the old 400 from the server matched the `code 400` branch and showed `alert.invalidImage`, so users got human copy, just the wrong copy. Mapped the error before aggregation, per the first of the two options suggested.

**Greptile (P1, chatThreadContainer)** — confirmed. `error` is only rendered by `ThreadMessageList.renderEmptyList()`, so in a thread with existing messages the recovery guidance was never shown. Switched to `Alert.alert`, matching how `profileEditContainer` surfaces the same condition. The pre-existing generic path has the same limitation, but it mattered more here because this branch exists specifically to tell the user how to recover.

**CodeRabbit (chatThreadContainer deps)** — confirmed, and introduced by this PR: `_handleAttachImage` had no `intl` usage before it, so the dependency list was correct until I added one. Added `intl`.

Full suite 660 passed / 1 skipped, typecheck clean, no new lint warnings.

PR #3363, opened 2026-07-22
feat(notifications): sync filter tabs and push toggles across all types

feruzm:
Companion to ecency/vision-next#1229, which does the same for web.

Notification types were exposed inconsistently on mobile:

| Type | Renders | Filter tab | Push toggle |
|---|---|---|---|
| replies, mentions, rvotes, follows | yes | yes | yes |
| reblogs, transfers, nfavorites, nbookmarks | yes | **no** | yes |
| scheduled_published | yes | **no** | yes |
| delegations | yes | **no** (enum only) | **no** |
| payouts, account_update, weekly_earnings | yes | **no** | **no** |

The filter bar offered 5 of the 10 enum values. This adds the missing filter entries and push toggles so every type the feed renders can also be filtered and muted.

## Notes

- `notifyTypesConst` values are enotify's `ACTIVITY_MAIN_TYPE_*` ints, validated server-side at device registration. Payouts is **19** — 16 is commented out upstream and must not be reused. The map is duplicated across `loginContainer`, `settingsContainer` and `applicationContainer`, so all three are updated together (note `applicationContainer` keys on the full `*Notification` name while the other two use the short name).
- The `account_updates` filter path is plural while the notification `type` string is `account_update` (singular). enotify routes on the plural.

## Filter labels moved to their own namespace

Labels previously came from `notification.<key>`, which for the newer types holds the notification **body** string — `payouts` is `"Payout received: ${amount}"` and `delegations` is `"delegated"`. Reusing those would have rendered template text in the dropdown, so labels now live under `notification.filters.*`.

## Defaults

Follows the existing convention rather than leaving the new toggles off: `initialState: true` plus a **v21 migration** backfilling the four keys, exactly as v20 did for `scheduledPublished`. `autoMergeLevel1` keeps persisted `notificationDetails` as-is on rehydration, so a new key is never defaulted from `initialState` without a migration.

**This means existing users start receiving delegation, payout, account-update and weekly-earnings pushes after upgrading.** That is deliberate and matches how every other mobile notification type behaves, but it is the reviewable decision in this PR.

## Verification

- `eslint` on all touched paths: 0 errors
- Full suite: 43 suites / 663 tests pass, including 3 new v21 migration tests (defaults ON when missing, preserves explicit false, no-op when notificationDetails absent)
- Both new filter paths return live data end to end: `/api/weekly_earnings/<user>` → `type: weekly_earnings`, `/api/account_updates/<user>` → `type: account_update`

The api-proxy allowlist entries these filters depend on were added separately; without them the new tabs would return empty.

## Summary by CodeRabbit

* **New Features**
  * Added new notification filters and settings for delegations, payouts, scheduled posts, account updates, and weekly earnings.
  * Added corresponding push notification toggles on the settings screen.
* **Bug Fixes**
  * Updated notification filter display text to use the correct localized labels.
  * Improved foreground notification handling for the new notification types to ensure consistent unread-activity updates and titles/bodies.
* **Chores**
  * Extended persisted notification preferences with a new default migration (new toggles default to enabled when missing) and added test coverage.


feruzm:
Triaged both P1s.

**Toggles skip the update path — confirmed and fixed** in da42a529ad. `_handleToggleChanged` switches on `actionType` and only forwarded the pre-existing `notification.*` cases, so all four new toggles hit `default`: no Redux dispatch, `_setPushToken` never called, switch snaps back. Added them to the case group.

**Translated labels lose their fallback — not reproducible, declining.** `src/index.tsx:30` builds the message table as an en-US underlay:

```js
messages={{ ...flattenMessages(messages['en-US']), ...flattenMessages(messages[locale]) }}
```

`flattenMessages` flattens to dot-notation keys, so `notification.filters.*` resolves from the en-US layer for every locale that has not translated it yet. Labels render in English rather than as `NOTIFICATION.FILTERS.REPLIES`. That underlay is deliberate and documented in the comment directly above the line. Same mechanism already covers `settings.notification.scheduled_published`, which is also en-US only today.

Crowdin picks the new keys up from en-US on the next sync, as with any other added string.

feruzm:
Triaged the CodeRabbit findings. Two of three actionable, one declined.

**Foreground + websocket allowlists — confirmed and fixed** in c144c4a28e. `payouts`, `account_update` and `weekly_earnings` were in neither allowlist, so they produced no foreground notification and no unread-count refresh.

Worth flagging that the suggested fix alone would have made things worse: the websocket→FCM `title` was a ternary chain ending in `` `@${source} delegated to @${target}` ``, so every newly-allowed type would have announced itself as a delegation. Replaced the chain with a switch that handles each allowlisted type explicitly, so an unhandled type can no longer inherit another type's wording.

**Route the four new toggle actions — already fixed** in da42a529ad, before this review landed.

**Bulk notification-settings reducer — declining.** `CHANGE_ALL_NOTIFICATION_SETTINGS` has exactly one dispatcher: the one-time legacy realm→redux settings migration in `migrationHelpers.ts:95`, gated by `setSettingsMigrated`. It is not an "enable/disable all" control — that path is `changeNotificationSettings({ type: 'notification' })`, which calls `_setPushToken` with the full `notifyTypes` list built from `notificationDetails`, and that list does include the four new keys.

The suggested diff is also behaviourally identical to the current code. Legacy realm payloads predate these fields, so `action.payload.X` is always `undefined`; with the `??` guard it resolves to `state.notificationDetails.X`, which is exactly what the existing `...state.notificationDetails` spread already yields. Adding the four lines changes no observable behaviour.

PR #3368, opened 2026-07-27
Fix video thumbnail generation and post editor passthrough

feruzm:
Video thumbnails are auto-generated from frames on mobile already, but three issues made it look like they were not.

### Frame generation (`speakUploaderModal`)

- A single `createThumbnail` rejection (unsupported codec, iOS `ph://` url) rejected the whole `showUploader` promise. The modal stayed open with an empty strip and the video uploaded with no thumbnail. Failures are now tolerated per frame.
- `react-native-image-crop-picker` omits `duration` on some Android paths, which made every `timeStamp` `NaN`. Falls back to a fixed duration.
- `availableThumbs` was not reset on open, so a previous video's frames could persist.

Frames are now sampled at the middle of each slice rather than at its start, so the first frame, commonly black, is no longer the implicit default. The default is the middle frame, and the header tile previews it so the user can see which thumbnail will be used.

### Post editor passthrough

`UploadsGalleryModal` ignored the `thumbnailUrl` argument from `onVideoUploaded`, and `postOptionsModal` hardcoded `videoThumbUrls={[]}`. A video only post therefore had no thumbnail candidate: the selection strip showed "add more images" and the post published with no `meta.image`.

The uploaded thumbnail url is now threaded up to `editorContainer` the same way `setIsUploading` is threaded down, and fed into the thumbnail selection list. The candidate list is deduped, and `videoThumbUrls` is included in the effect dependencies since the url arrives after the body update.

Waves already handled this correctly and is unchanged.

### Verification

- `npx tsc --noEmit` clean on changed files
- `npx eslint` 0 errors, no new warnings
- `npx prettier --check` clean
- Jest: 663 passed, 1 skipped

Not exercised on a device yet, the upload path needs a real 3Speak account.

## Summary by CodeRabbit

* **New Features**
  * Added video thumbnail generation that samples multiple frames and selects a sensible default.
  * Added support for carrying multiple pre-supplied video thumbnail URLs through the post thumbnail selection dialog.
  * Video thumbs captured from uploads can now be passed into the editor when saving or editing posts.
* **Bug Fixes**
  * Avoided invalid timestamp calculations during thumbnail generation.
  * Prevented stale thumbnail results from overwriting newer uploads.
  * Upload button is disabled while thumbnails are generating, and partial thumbnail failures are tolerated.
  * Thumbnail options are deduplicated and only retained while their embed URLs remain in the post body.

### DIFF
--- a/src/components/markdownEditor/children/editorToolbar.tsx
+++ b/src/components/markdownEditor/children/editorToolbar.tsx
@@ -1,5 +1,6 @@
 import React, { useEffect, useRef, useState } from 'react';
 import {
+  Alert,
   AppState,
   Keyboard,
   View,
@@ -43,6 +44,7 @@ import ROUTES from '../../../constants/routeNames';
 import { DEFAULT_USER_DRAFT_ID } from '../../../redux/constants/constants';
 import { TextFormatModal } from './textFormatModal';
 import { selectCurrentAccount } from '../../../redux/selectors';
+import { hasThreeSpeakEmbed } from '../../../providers/speak/beneficiary';
 
 // Per-account session dismissals for the quest chip; once closed it stays
 // hidden for that account until the app restarts.
@@ -60,6 +62,8 @@ type Props = {
   setIsUploading: (isUploading: boolean) => void;
   handleMediaInsert: (data: MediaInsertData[]) => void;
   handleVideoThumb?: (embedUrl: string, thumbUrl: string) => void;
+  /** Reads the live body. `postBody` is only as fresh as the last render. */
+  getPostBody?: () => string;
   handleOnAddLinkPress: () => void;
   handleOnClearPress: () => void;
   handleOnMarkupButtonPress: (item) => void;
@@ -81,6 +85,7 @@ export const EditorToolbar = ({
   setIsUploading,
   handleMediaInsert,
   handleVideoThumb,
+  getPostBody,
   handleAiToolUsed,
   handleOnAddLinkPress,
   handleOnClearPress,
@@ -247,6 +252,15 @@ export const EditorToolbar = ({
   };
 
   const _showVideoUploads = () => {
+    // One 3Speak video per post, matching web. The threespeakfund beneficiary is a single
+    // flat share on the post, so a second video would publish without its own payout route.
+    if (hasThreeSpeakEmbed(getPostBody?.() ?? postBody ?? '')) {
+      Alert.alert(
+        intl.formatMessage({ id: 'alert.notice' }),
+        intl.formatMessage({ id: 'video-upload.error-one-per-post' }),
+      );
+      return;
+    }
     _showUploadsExtension(Modes.MODE_VIDEO);
   };
 
--- a/src/components/markdownEditor/children/editorToolbar.tsx
+++ b/src/components/markdownEditor/children/editorToolbar.tsx
@@ -59,6 +59,7 @@ type Props = {
   suggestedPrompt?: string;
   setIsUploading: (isUploading: boolean) => void;
   handleMediaInsert: (data: MediaInsertData[]) => void;
+  handleVideoThumbUrl?: (url: string) => void;
   handleOnAddLinkPress: () => void;
   handleOnClearPress: () => void;
   handleOnMarkupButtonPress: (item) => void;
@@ -79,6 +80,7 @@ export const EditorToolbar = ({
   suggestedPrompt,
   setIsUploading,
   handleMediaInsert,
+  handleVideoThumbUrl,
   handleAiToolUsed,
   handleOnAddLinkPress,
   handleOnClearPress,
@@ -355,6 +357,7 @@ export const EditorToolbar = ({
               username={currentAccount.name}
               hideToolbarExtension={_hideExtension}
               handleMediaInsert={handleMediaInsert}
+              onVideoThumbUrl={handleVideoThumbUrl}
               setIsUploading={setIsUploading}
             />
             <TextFormatModal
--- a/src/components/markdownEditor/children/editorToolbar.tsx
+++ b/src/components/markdownEditor/children/editorToolbar.tsx
@@ -59,7 +59,7 @@ type Props = {
   suggestedPrompt?: string;
   setIsUploading: (isUploading: boolean) => void;
   handleMediaInsert: (data: MediaInsertData[]) => void;
-  handleVideoThumbUrl?: (url: string) => void;
+  handleVideoThumb?: (embedUrl: string, thumbUrl: string) => void;
   handleOnAddLinkPress: () => void;
   handleOnClearPress: () => void;
   handleOnMarkupButtonPress: (item) => void;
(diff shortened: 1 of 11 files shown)


feruzm:
Review triage, all pushed in `26c2d60`.

**Codex P1, video thumbnails never reach `meta.image`** Confirmed and fixed. This was the important one: `extractMetadata` builds `out.image` from body images plus `videoThumbUrls`, and uses `thumbUrl` only to reorder that list. With `videoThumbUrls` omitted, `matchedImages` stayed empty for a video only post, so `out.image` was never set and the previous commit only fixed the picker, not the published post. All four paths that build post metadata now pass the list: publish, edit, draft save and template save.

**Greptile P1 and CodeRabbit, removed videos retain thumbnails** Confirmed and fixed. Thumbnails are now stored as `{ embedUrl, thumbUrl }` and filtered through a new `filterActiveVideoThumbs` helper against the current body, so deleting the embed drops its thumbnail. This also covers CodeRabbit's separate point about resetting on clear: an empty body yields an empty candidate list, so no extra reset hook is needed.

**CodeRabbit, generation race and upload enabled before frames are ready** Confirmed and fixed with a generation id, results only committed when still current, and the upload button disabled while generating.

**CodeRabbit, `thumbUrl` missing from effect dependencies** Fixed.

Added unit tests for `filterActiveVideoThumbs` covering removal, multiple videos and clear.

Verification: eslint 0 errors, prettier clean, Jest 668 passed. Note `npx tsc` cannot run in this repo at all, base included, because `tsconfig.json` sets `moduleResolution: node` which conflicts with `customConditions` from the extended React Native config. Forcing `--moduleResolution bundler` produces a few hundred pre-existing errors across these files. I compared error counts against the base commit and the only additions are the same untyped-props class already present throughout these components.

Still not exercised on a device.

feruzm:
Re-review triage. Greptile's original P1 (removed videos retain thumbnails) is resolved and it has dropped that finding; the 3/5 is a new, different issue.

**Greptile P1, draft reload loses video thumbnails** Confirmed and fixed. Traced it: reopening a draft restores `thumbUrl` from `draft.meta.image[0]`, but `videoThumbs` starts empty, so the candidate list came out empty and `thumbSelectionContent` then called `onThumbSelection('')` and cleared `thumbUrl` outright. The cover vanished from both selection and published `meta.image`. Saved images are now kept as `restoredThumbUrls` and offered while the body still holds a 3Speak embed. This is the same gap Codex raised on the web PR, which I had fixed there but not here.

**CodeRabbit, `body.includes(embedUrl)` matches arbitrary substrings** Correct. Videos are inserted as a raw url on its own line (`applyMediaLink` uses `\n${value}\n` for `MODE_VIDEO`), so exact matching is straightforward. `filterActiveVideoThumbs` is replaced by `collectVideoThumbUrls`, which compares against `extractUrls(body)` exactly. Added the regression test for the `/aaa` inside `/aaa2` collision.

**CodeRabbit, persist video thumbnails in local draft autosave** No separate change needed, and worth explaining. `_saveCurrentDraft` writes `meta.image = [thumbUrl]`, which was only losing the cover because `thumbUrl` was being cleared by the bug above. With that fixed, the autosave stores the thumbnail and `restoredThumbUrls` brings it back on reopen, so the loop closes without merging the list into `meta.image` separately.

One thing the helper now also does: it drops thumbnails that are already images in the body. `extractMetadata` concatenates body images with `videoThumbUrls` without deduping, so without this a thumbnail appearing in both would land in `meta.image` twice.

Verification: eslint 0 errors, prettier clean, Jest 672 passed (9 new). Still not exercised on a device.

feruzm:
Second re-review triage, pushed in `def904d`.

Greptile is right again, and its new finding exposed a flaw in my previous fix rather than repeating the old one.

Restoring the draft cover as a flat list of urls, admitted whenever the body held any 3Speak url, fixed the single video case but not a multi video draft: an unassociated url cannot be tied to the video it came from, so deleting one video left its cover selectable and eligible for `meta.image`. Worth noting the obvious tightening does not work either. Gating on "exactly one embed in the body" still fails, because deleting one video from a two video draft brings the count back to one and the stale cover is admitted anyway. I caught that while writing the test for it.

The fix is to rebuild the association at restore time instead of carrying a bare list. `restoreVideoThumbs(draft.body, draft.meta.image)` returns `{ embedUrl, thumbUrl }` when the draft holds exactly one embed, and that pair then flows through the same exact match filter as a session upload. Deleting that video drops its cover, and a replacement video does not inherit it. A draft with several embeds cannot be attributed, so it restores nothing rather than guessing, which is the behaviour before any of this was carried at all.

Doing it this way also removed the separate `restoredThumbUrls` state and the special case in `collectVideoThumbUrls`, so restored and session thumbnails now take exactly the same path.

On the missing coverage Greptile flagged in `editor.test.ts`: added tests for attribution, refusal to guess with multiple embeds, removal of the restored video, and replacement by a different video.

Remaining limitation, stated plainly: a reopened multi video draft offers no restored cover. Fixing that needs the associations persisted in the draft, which is a draft schema change I did not want to make unilaterally, or a 3Speak lookup per embed, which needs an endpoint the mobile provider does not currently have (`speak.ts` exposes `setVideoThumbnail` but no getter). Happy to do either if you prefer.

Verification: eslint 0 errors, prettier clean, Jest 675 passed. Still not exercised on a device.

feruzm:
On Greptile's remaining finding (multi-video drafts lose thumbnail state): I checked whether that state is reachable, and whether it should be.

It is reachable on mobile today. `_showVideoUploads` never inspected the body, so the video button could be tapped again to add a second 3Speak embed. Web does not allow it: `onVideoUploaded` becomes `undefined` once `hasThreeSpeakEmbed(body)` is true.

It also should not be reachable, for a reason beyond thumbnails. Mobile already carries `hasThreeSpeakEmbed` in `providers/speak/beneficiary.ts`, where it gates a single flat 11% `threespeakfund` route on the post. A post with two videos pays one 11% share, so the second video publishes with no payout route of its own. That is a pre-existing bug independent of this PR, and worse than a missing thumbnail since it misroutes 3Speak revenue.

So rather than build association persistence for a state that should not exist, `cb376f1` brings mobile in line with web and allows one video per post. Greptile's case is now unreachable: a post can only ever hold one embed, which is exactly the shape `restoreVideoThumbs` can attribute.

Two implementation notes:

- The guard reads the body through a getter, not the `postBody` prop. The editor is deliberately uncontrolled and `_setTextAndSelection` writes straight to native via `setNativeProps` without a re-render, so `postBody` is only as fresh as the last render and would have made the guard unreliable.
- `extractVideoEmbedUrls` now reuses `hasThreeSpeakEmbed` rather than its own looser pattern, so what counts as a video for thumbnails cannot drift from what counts as one when the beneficiary is attached. If that regex ever misses a 3Speak url format, both fail together instead of one allowing a video the other leaves unpaid.

New string `video-upload.error-one-per-post` in en-US.json only, for Crowdin to pick up.

This is a UX behaviour change on mobile, so flagging it plainly: users who could previously add two videos to one post no longer can. Happy to revert that part if you would rather keep the current behaviour and accept the draft limitation instead.

Verification: eslint 0 errors on changed code files (the 2 errors reported in en-US.json are pre-existing `${amount}` placeholders at lines 436-437, unchanged on base), prettier clean, Jest 675 passed. Waves is unaffected, it uses its own uploader and already allowed one video. Still not exercised on a device.

Issue #46, opened 2018-10-04
Splash screen design.


ue:
Need animation or UI (sketch) design.

PR #3378, opened 2026-07-31
fix(points): label burned points, correct the microphone usage string

feruzm:
Two small fixes that are already visible in production, split out from the dictation feature so they are not blocked behind it.

## Burned points had no label

AI spends now burn instead of transferring to the treasury (ePoints #46, deployed). A burn row carries **no counterparty at all** — `sender` and `receiver` are both null, which is exactly what distinguishes it from a treasury transfer — so it reports type **997**.

Mobile has no entry for 997, so `POINTS[type] || POINTS.default` renders every AI spend as a generic "Activity" row with no indication of what was spent or on what. That is live right now for ai-assist and ai-image.

Added the entry to **both** copies of the map. `src/utils/points.ts` is an unused byte-identical duplicate of `src/constants/options/points.ts` (only the latter is imported anywhere), so updating just the live one would have quietly forked them. The duplication is worth its own cleanup, but deleting a file is not this PR's job.

Note a refunded burn arrives as the same type with a positive amount. Mobile already renders amount signs, so it reads correctly without a separate entry.

## The microphone usage string described the camera

`NSMicrophoneUsageDescription` read: *"Ecency requires access to your camera to let you insert recordings and videos to content you are creating."*

App Review reads these strings, and so do users at the moment they decide whether to grant. It also matters more shortly, since dictation puts this prompt in front of everyone who tries it.

## Verification

688 mobile tests passing, eslint clean on the changed files, i18n JSON validated.

## Not in this PR

The dictation feature itself (recorder, sheet, toolbar button, `RECORD_AUDIO`). It needs a new native dependency and the SDK at 2.3.70, and I cannot build or run this app in my environment — see the PR discussion.

## Summary by CodeRabbit

* **New Features**
  * Added wallet activity tracking for points spent on AI features.
  * Added support for displaying burned-point activity with a fire icon and point value.
  * Added refund activity for points returned after unsuccessful AI requests.

* **Bug Fixes**
  * Improved the iOS microphone permission description to accurately reflect dictation and audio recording.


feruzm:
Fixed. Codex, CodeRabbit and greptile all ran; one actionable finding between them, plus something greptile spotted in its summary that was also worth fixing.

## P2 — refunded burns were labelled as spending

Correct, and I should own this one specifically: **I asserted in the PR description that the amount sign made it read correctly.** It does not. The sign is rendered, but the label is not derived from it, so a refund showed "Spent on AI" beside a credit. I had already handled exactly this on web (the component passes a direction) and then hand-waved it here.

`resolvePointType` now owns the lookup and both call sites use it — the wallet query and the points container — rather than the sign check being duplicated in two places that could drift.

I also reworded the spend label to **"AI usage"**, so it reads sensibly whichever way the entry is reached, with a separate "AI refund" for the credit.

Only 997 needs this. Transfers carry a counterparty that already states direction, which the test pins.

## Locale diff noise

greptile's summary noted the PR "replaces escaped Unicode punctuation with equivalent literal characters". That was unintentional: I added the keys with a `json.load`/`dump` round-trip, which rewrote escaping across the whole file and un-escaped six unrelated bullet and ellipsis strings.

Functionally identical, but it is noise in a file Crowdin owns. Redone as a targeted text insert, so the locale diff is now exactly the four added keys.

## Verification

**694 mobile tests passing** (688 + 6 new), eslint clean on every touched file, locale JSON validated.

The new tests cover both signs, numeric and string amounts, other types being left alone, unknown types falling back, and malformed items not throwing.

## Still not in this PR

The dictation feature itself. It needs a new native dependency and I cannot build this app in my environment (no Java, and iOS builds are not possible on Linux), so I would be shipping native-integrated code that has never been compiled.

PR #3386, opened 2026-07-31
feat(editor): dictate posts from the mobile editor toolbar

feruzm:
> **Draft: this has never been compiled or run.** I cannot build this app in my environment (no Java, and iOS builds are not possible on Linux), so everything below is verified at the JS layer only. Expect the first real build to need fixes.

Adds a microphone button beside the video upload, opening a sheet that records, shows the running cost, and inserts the transcript at the caret.

Final piece of the dictation series. ePoints (#45/#46/#47), vision-api (#62) and vision-next (#1285/#1286) are merged and live.

## Written against the real API, not a guessed one

I unpacked the published `expo-audio@0.4.9` tarball and read its types rather than writing from memory: `useAudioRecorder`, `useAudioRecorderState` (which is where `durationMillis` comes from), `requestRecordingPermissionsAsync`, `setAudioModeAsync`, `RecordingPresets`.

The version is what Expo SDK 53 pins in `bundledNativeModules.json` (`~0.4.9`), not something I picked.

`HIGH_QUALITY` records AAC in `.m4a` on both platforms, which the backend accepts directly — no transcode, and no dependence on whichever container a given OS version prefers.

## Carries over what web review found

The web implementation took several review rounds. Rather than repeat those mistakes here, this starts from their conclusions:

- **Duration rounds UP.** The server bills whole 30s units off the real duration, so flooring quotes one unit for a clip that costs two.
- **Recording stops a second before the cap.** Stopping exactly at it lets the real duration drift past by the time the recorder flushes, and the server then rejects an upload the user already waited through.
- **A permission prompt answered after the sheet closed cannot start a recording.** A generation counter is checked after every `await` in `start()` — otherwise a late grant opens a microphone with no UI able to close it.
- **The sheet cannot be dismissed while transcribing.** Points are charged the moment the request lands, so closing mid-flight would pay for a transcript and throw it away.
- **Unmount marks it closed**, so a late response cannot edit the draft of a screen the user has left.
- **The idempotency key is held across retries**, so a retry after a lost response replays rather than paying twice.
- **Pricing is never guessed.** Recording is blocked until the price endpoint answers, because the user pays the server's number, not ours.

## Placement

The mic sits next to the video upload rather than among the AI text actions. Dictation is media capture, so it belongs with the other capture controls.

## Insertion

Dictated text goes in at the caret via `applySnippet`, not replacing the body the way the AI assist edit actions do. Someone dictating mid-draft is adding to it.

## What is verified

- `tsc` clean on all new code (the one `tsconfig.json customConditions` error is pre-existing on `development`)
- eslint 0 errors on every touched file
- 708 tests passing
- `expo-audio` mocked in `jest.setup.ts` alongside the other native modules, so a future test reaching this sheet does not fail on the import
- Lockfile change is 5 lines, only `expo-audio`

## What is NOT verified

Everything native. Specifically worth attention on first build:

1. **`pod install`** for the new native module, and the iOS build generally.
2. **Android runtime permission.** `RECORD_AUDIO` is declared in the manifest, but Android 6+ also needs a runtime grant — `requestRecordingPermissionsAsync` should handle it, though that path is untested.
3. **The upload shape.** `useAiTranscribe` builds a `FormData` and appends `params.audio`. On web that is a `Blob`; here it is React Native's `{ uri, name, type }` file object, which RN's `FormData` understands but TypeScript does not, so it is cast. **This is the single most likely thing to be wrong on first run.**
4. **iOS audio session.** `setAudioModeAsync({ allowsRecording: true, playsInSilentMode: true })` is set before recording and reset on unmount; whether that interacts well with other audio in the app is untested.

## Summary by CodeRabbit

* **New Features**
  * Added voice dictation to the editor through a microphone action.
  * Record audio, view duration and pricing, and submit recordings for transcription.
  * Insert transcribed text at the current cursor position.
  * Added handling for permissions, recording limits, retries, and transcription errors.
  * Added localized messaging for dictation controls and status updates.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -4,6 +4,7 @@
 
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.CAMERA"/>
+    <uses-permission android:name="android.permission.RECORD_AUDIO" />
     <uses-permission android:name="com.android.vending.BILLING" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.DOWNLOAD_WITHOUT_NOTIFICATION" />
--- a/jest.setup.ts
+++ b/jest.setup.ts
@@ -49,6 +49,28 @@ jest.mock('@react-native-async-storage/async-storage', () =>
 );
 
 // Mock Expo crypto
+// expo-audio is a native module: importing it unmocked fails any test that reaches
+// the dictation sheet. Mirrors the shape useDictationRecorder actually consumes.
+jest.mock('expo-audio', () => ({
+  RecordingPresets: { HIGH_QUALITY: {} },
+  requestRecordingPermissionsAsync: jest.fn(async () => ({ granted: true })),
+  setAudioModeAsync: jest.fn(async () => undefined),
+  useAudioRecorder: jest.fn(() => ({
+    isRecording: false,
+    uri: null,
+    record: jest.fn(),
+    stop: jest.fn(async () => undefined),
+    prepareToRecordAsync: jest.fn(async () => undefined),
+  })),
+  useAudioRecorderState: jest.fn(() => ({
+    canRecord: true,
+    isRecording: false,
+    durationMillis: 0,
+    mediaServicesDidReset: false,
+    url: null,
+  })),
+}));
+
 jest.mock('expo-crypto', () => ({
   digestStringAsync: jest.fn(),
 }));
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "domain-browser": "^5.7.0",
     "events": "^1.0.0",
     "expo": "^53.0.20",
+    "expo-audio": "~0.4.9",
     "expo-crypto": "~14.1.5",
     "expo-iap": "^4.6.0",
     "expo-image": "~2.4.0",
--- a/src/components/dictationModal/dictationModal.styles.ts
+++ b/src/components/dictationModal/dictationModal.styles.ts
@@ -1,0 +1,91 @@
+import { ViewStyle, TextStyle } from 'react-native';
+import EStyleSheet from 'react-native-extended-stylesheet';
+
+export default EStyleSheet.create({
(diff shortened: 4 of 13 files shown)


feruzm:
Both findings fixed. All three bots ran (greptile 4 comments, codex 2, coderabbit 1) and they converged on the same two rejection paths.

## Both were the same omission in two places

I handled the *expected* failure — a denied permission prompt — and treated the native calls around it as if they could not reject. They can.

**start()**: `requestRecordingPermissionsAsync`, `setAudioModeAsync` and `prepareToRecordAsync` can all reject outright, and an audio-service interruption does exactly that. The hook then sat in `requesting` forever: Record disabled, an unhandled rejection logged, no way out but closing and reopening the sheet.

**stop()**: a rejected stop returned before setting the result or leaving `recording`, so the stop button stayed up but no longer did anything, and there was no recording to submit either.

## The fix

A new `failed` state, kept **distinct from `denied`** because nothing was refused — the hardware or session failed. So the sheet says "could not start recording" and leaves Record enabled to retry, rather than sending the user to Settings for a permission that is already granted.

The superseded-generation check applies inside the `catch` too, so a rejection from an abandoned attempt cannot clobber the recording that replaced it. Same reasoning as the denial path, which already had that guard — this is the third place that check turned out to be needed.

Also caught the floated `recorder.stop()` calls in `reset()`, on unmount, and behind the duration cap. Those cannot be awaited, so an unhandled rejection there surfaced as a redbox over a teardown the user cannot act on anyway.

## Status update on the draft warning

**Android Build passed (39m21s)**, so the "never compiled" caveat in the description is now out of date for Android — CI has built it even though I could not. iOS remains unbuilt, and nothing has been run against a real microphone.

`Analyze (actions)` fails at "Initialize CodeQL", which is a CodeQL setup failure. This PR touches no workflow files, so it is unrelated.

## Verification

`tsc` clean, eslint 0 errors, 708 tests passing.

Still unverified: iOS build, and real recording on a device. The upload shape remains the most likely runtime surprise — `useAiTranscribe` appends `params.audio` to `FormData`, which is a `Blob` on web and React Native's `{ uri, name, type }` object here.

PR #3401, opened 2026-08-01
fix(dictation): stop polling a native audio object, and show what was recorded

feruzm:
Fixes the iOS TestFlight crash reported after recording and pressing Insert, plus the UX gaps around a stopped recording.

## The crash

Sentry event (`EXC_BAD_ACCESS`, `KERN_INVALID_ADDRESS`, iOS 26.6, iPhone17,1). The breadcrumbs are decisive:

```
[fetch]   POST /private-api/ai-transcribe          ← completed
[console] FunctionCallException: Calling the 'get' function has failed
          (at ExpoModulesCore/SyncFunction...)
          → EXC_BAD_ACCESS
```

**The upload worked.** The crash is in expo-audio afterwards.

`useAudioRecorderState` polls the recorder's **synchronous native `getStatus()`** on a `setInterval` — I had set it to 250ms, twice the default. And **sheets in this app stay mounted** (CLAUDE.md says so explicitly), so that poll never stops. It keeps calling into a native audio object for the rest of the session, and once that object is no longer valid — after stop, after `setAudioModeAsync({allowsRecording:false})`, or when iOS reclaims the audio session — the sync getter faults and takes the process down.

Duration is now timed from the wall clock, and `recorder.uri` is read once at stop. There is no recurring native call left.

Pleasing convergence: the web implementation already times from the clock, for a completely unrelated reason (Chrome's WebM blobs report `Infinity` when decoded). Both platforms now agree that the recorder is not the place to ask what time it is.

## The same wrong assumption caused three more bugs

I wrote the component as if it unmounts. It does not.

- **The timer snapped to 0:00 on stop** — the recorder zeroes its own duration, and that was what the clock rendered. Reported alongside the crash. The measured duration is now kept, so a finished recording shows its real length.
- **State persisted across opens.** An old recording sat there ready to be inserted into a *different* draft. Now reset per open keyed on `payload`, which is this repo's documented convention for sheets.
- **The unmount cleanup never ran** — so none of the teardown I wrote was doing anything.

## Stopped-recording UX

Also reported: after stopping, nothing told the user what state they were in, and "Record again" silently discarded audio they had not inserted yet.

- The sheet now says the clip is recorded and awaiting insertion.
- A caption warns that recording again discards it, shown before the destructive action rather than after.

## Verification

`tsc` clean, eslint 0 errors, 732 tests passing.

Confirmed by grep that no `useAudioRecorderState` import or call site remains — the only matches are in the comment explaining why it is avoided, which is deliberate so this is not reintroduced.

Not verified: a real iOS device. The crash was found from the Sentry event rather than reproduced locally, since I cannot build iOS. Worth a TestFlight pass on the same flow that crashed — record, stop, insert, then leave the app open for a minute.

## Summary by CodeRabbit

- **New Features**
  - Added a recorded-status indicator after dictation stops.
  - Added a warning that recording again will discard the current recording.
  - Recording duration now remains visible after stopping.

- **Bug Fixes**
  - Opening a new dictation sheet now starts with a clean recording state.
  - Improved recording start and stop reliability, including handling timing limits and cleanup.
  - Prevented recordings and related state from persisting between modal openings.


feruzm:
Both bots were right, and the flow has changed since — one push covering both.

## The compile break

Correct on both counts: `setServerRejectedForPoints` and `setPendingCost` were referenced but never declared. Those setters are from the **web** dialog, not this one; they got in because I patched this component with blind string replaces against a file I misremembered.

I hit the same wall myself while adding the new flow and rewrote the component properly rather than patching further, which removed them. Verified: no references remain and `tsc` is clean.

The lesson is the one the bug demonstrates — I should have read the file before editing it, which is exactly what eventually found the problem.

## New in this push: transcribe on stop

Requested by the user, and it collapses record-stop-review-insert into record-stop:

- Transcription fires when recording stops; the text goes straight into the post.
- The sheet stays open with **Record more**, so dictating several paragraphs no longer means reopening it each time.
- **Done** closes the sheet; closing is no longer a side effect of inserting.

### It would have shipped broken without a second fix

Consecutive inserts overwrote each other. `applySnippet` leaves the inserted text **selected** (`{start, start + len}`), and `replaceBetween` replaces whatever is selected — so segment two replaced segment one instead of following it.

Dictation now inserts through its own path that leaves the caret **collapsed after** the text, with a single space between segments unless the caret already follows whitespace.

### Consequences of transcribing on stop, handled

- **Stopping is now the moment Points are spent**, with no separate confirmation. The running cost is already shown throughout recording, and the hint now states that stopping transcribes and adds to the post. Worth a second opinion on whether that is enough consent for a paid action.
- A failure keeps the audio **and** the idempotency key, so *Try again* reuses the recording rather than making the user repeat themselves — and a request that actually landed replays instead of charging twice.
- The auto-submit effect is keyed on the recording's `uri`, so a re-render or retry cannot fire a second paid request for the same audio.
- An empty transcript still reports and resets, since silence is charged.

Also dropped the now-unused `dictation.max/rerecord/insert/recorded` strings rather than leaving them for Crowdin to translate.

## Verification

`tsc` clean, eslint 0 errors, 732 tests passing.

Still not verified on a device — the multi-segment insert ordering in particular is worth a TestFlight pass, since that is the part that was silently broken.

PR #3467, opened 2026-08-06
fix(dictation): stop calling into the recorder after expo-audio released it

### LINKED ISSUE #3466: Dictation sheet: tapping Done crashes the app (native call on a released expo-audio recorder)
## Symptom

Open the editor toolbar dictation sheet, record, then tap **Done**. The sheet closes and the app crashes.

## Root cause

The dictation recorder's unmount cleanup calls into the `expo-audio` recorder after expo has already released it.

1. **Done** runs `close()` (`src/components/dictationModal/dictationModal.tsx:77-82`), which calls `SheetManager.hide(SheetNames.DICTATION)`.
2. `react-native-actions-sheet` 0.9.7 **unmounts** the sheet on hide. `RenderSheet` renders `!visible ? null : <Sheet/>` (`dist/src/provider.js:154`) and `visible` flips to false when the hide animation publishes `onclose_<id>` (`dist/src/index.js:408`).
3. React runs effect cleanups in hook-declaration order. `useAudioRecorder` is the first hook in `useDictationRecorder` (`src/hooks/useDictationRecorder.ts:45`), and its internal `useReleasingSharedObject` registers the first cleanup, which calls `recorder.release()`. That detaches the JS object from its native counterpart.
4. Our own unmount cleanup (`src/hooks/useDictationRecorder.ts:166-178`) runs next and reads `recorder.isRecording` on line 172. That is a native property getter on a released shared object, so it throws `NativeSharedObjectNotFoundException` on iOS (`expo-modules-core/ios/Core/DynamicTypes/DynamicSharedObjectType.swift:97`) or `InvalidSharedObjectIdException` on Android (`SharedObjectRegistry.kt:96-99`).
5. The throw lands in a commit-phase cleanup with **no error boundary above it**. `SheetProvider` renders registered sheets as siblings of `<Application/>` (`src/index.tsx:39-43`), while `ErrorBoundary` lives inside `Application` (`src/screens/application/index.tsx:17`). It reaches RN's global handler as a fatal.

## The "sheets stay mounted" assumption is wrong

`src/hooks/useDictationRecorder.ts:164-165` says "Sheets in this app stay mounted, so this rarely fires", and the Sheets section of `CLAUDE.md` says the same. In 0.9.7 the sheet unmounts on every hide, so that cleanup runs on every close. Verified with a throwaway jest test driving the real `SheetProvider` + `SheetManager`: a probe sheet logged `["MOUNT"]` after show and `["MOUNT", "UNMOUNT"]` after hide.

A second throwaway test rendered `useDictationRecorder` against a mock mirroring `expo-audio`'s real hook ordering. Cleanup order was `release()` first, then the `isRecording` read with `released=true`, and the error escaped `unmount()` uncaught.

Not specific to having recorded: it should also reproduce on backdrop tap, back press, and opening the sheet then closing it without recording at all.

## Knock-on effect

Because line 172 throws, line 175's `setAudioModeAsync({ allowsRecording: false })` never runs, so on iOS the session is left active in `.playAndRecord` after dictation.

## Fix

The cleanup does not need to touch the recorder. Expo already does that teardown during release: iOS `sharedObjectWillRelease()` stops a running recorder, Android `sharedObjectDidRelease()` calls `reset()` which releases the `MediaRecorder`.

- Drop the recorder access from the unmount cleanup and leave it clearing the interval.
- Move the audio-mode reset to where the recorder is still alive (`close()` / `onClose`).
- Add a regression test that unmounts the hook with a released recorder.
- Correct the stale "sheets stay mounted" claim in `CLAUDE.md` and in the sheet comments written for it.

`useDictationRecorder` is the only place in the app using an Expo shared-object hook, so the blast radius stops there.

feruzm:
Closes #3466

Tapping **Done** in the dictation sheet closed the sheet and killed the app.

## What was happening

Registered sheets unmount on hide (`SheetProvider` renders `!visible ? null : <Sheet/>`), so `useDictationRecorder`'s unmount cleanup runs on every close, not "rarely" as its comment claimed.

`useAudioRecorder` is the first hook in `useDictationRecorder`, so its internal `useReleasingSharedObject` cleanup runs first and calls `recorder.release()`, detaching the JS object from its native counterpart. The cleanup declared after it then read `recorder.isRecording`, which throws `NativeSharedObjectNotFoundException` on iOS and `InvalidSharedObjectIdException` on Android.

Sheets render as siblings of `<Application/>` (`src/index.tsx`), so they sit outside the app's `ErrorBoundary`. The throw reached React Native's global handler as a fatal.

## Change

- The unmount cleanup no longer touches `recorder`. It clears the ticker and deactivates the recording session, nothing else.
- Dropping the `recorder.stop()` leaks nothing: `release()` already stops a running recorder on both platforms (iOS `sharedObjectWillRelease`, Android `sharedObjectDidRelease` -> `reset`).
- Side benefit: `setAudioModeAsync({ allowsRecording: false })` used to be skipped because the throw happened on the line above it, leaving the iOS session active in `.playAndRecord` after dictation. It now runs.
- Corrects the "sheets stay mounted" claim in `CLAUDE.md` and the comments written against it, and documents that sheets sit outside the `ErrorBoundary`.

## Test plan

- New `src/hooks/useDictationRecorder.test.tsx` mounts the hook against a mock mirroring expo-audio's real hook ordering, where the recorder throws on any member access after release. Verified it **fails** on the pre-fix code with `Unable to find the native shared object associated with given JavaScript object (isRecording)` and passes after.
- `yarn test:ci` green: 738 passed, 1 skipped.
- `tsc --noEmit` clean, eslint clean on changed files.
- Device check worth doing: dictate, tap Done, and confirm no crash; then confirm the same for backdrop tap, back press, and opening then closing without recording, which all took the same path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dictation cleanup when closing the recording modal.
  * Prevented crashes caused by accessing released recording resources during unmount.
  * Ensured recording sessions are safely deactivated and pending status checks are cancelled.

* **Documentation**
  * Clarified modal mounting, unmounting, state reset, error handling, and native-resource cleanup behavior.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -172,7 +172,13 @@ Uses `react-native-actions-sheet` for globally-registered bottom sheets:
 - Registry: `src/navigation/sheets.tsx` — all sheets registered with `SheetNames` enum
 - Show from anywhere: `SheetManager.show(SheetNames.MY_SHEET, { payload: {...} })`
 - Returns value: `const result = await SheetManager.show(...)` — resolves when sheet closes
-- Sheets stay mounted — reset state in `useEffect(() => { ... }, [payload])`
+- Sheets are registered globally but MOUNT on show and UNMOUNT on hide (`SheetProvider`
+  renders `!visible ? null : <Sheet/>`), so mount-time state resets are enough and any
+  unmount cleanup runs on every close, not rarely
+- Sheets render as siblings of `<Application/>` (`src/index.tsx`), so they sit OUTSIDE the
+  app's `ErrorBoundary`: a throw from a sheet's render or effect cleanup is uncaught and
+  fatal. Be careful with native/Expo shared objects in cleanups. Expo hooks such as
+  `useAudioRecorder` release their native object before any cleanup you declare after them
 
 ### Styling
 
--- a/src/components/dictationModal/dictationModal.tsx
+++ b/src/components/dictationModal/dictationModal.tsx
@@ -81,10 +81,9 @@ export const DictationModal = ({ payload }: SheetProps<SheetNames.DICTATION>) =>
     SheetManager.hide(SheetNames.DICTATION);
   }, [reset]);
 
-  // Sheets in this app STAY MOUNTED, so the unmount cleanup below effectively never
-  // fires and state would otherwise persist into the next open -- stale segment
-  // counts, or a failed recording offered for retry against a different draft.
-  // Reset per open, which is the documented convention for sheets here.
+  // Reset per open: stale segment counts, or a failed recording offered for retry
+  // against a different draft, must never survive into the next presentation.
+  // The sheet unmounts on hide, so this runs on mount for every presentation.
   useEffect(() => {
     closedRef.current = false;
     reset();
--- a/src/hooks/useDictationRecorder.test.tsx
+++ b/src/hooks/useDictationRecorder.test.tsx
@@ -1,0 +1,83 @@
+import React from 'react';
+import TestRenderer, { act } from 'react-test-renderer';
+
+// Mirrors expo-audio for real: useAudioRecorder delegates to useReleasingSharedObject,
+// which registers its release() cleanup FIRST, so it runs before any cleanup declared
+// later in useDictationRecorder. After release the JS object is detached from its
+// native counterpart and every native member access throws.
+const mockState = { released: false };
+const mockSetAudioModeAsync = jest.fn(async () => undefined);
+
+function mockThrowIfReleased(member: string) {
+  if (mockState.released) {
+    throw new Error(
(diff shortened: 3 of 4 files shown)


feruzm:
Re the `react-test-renderer` nitpick: declining it. `react-test-renderer` is the repo's only component-test renderer. There is no `@testing-library/react-native` in `package.json` or `node_modules`, and the two existing component tests (`src/components/postCard/children/postCardContent.test.tsx`, `src/components/autoHeightImage/autoHeightImage.test.tsx`) open with the identical `import TestRenderer from 'react-test-renderer'`. So this follows the established pattern rather than introducing a new one, and there is nothing to migrate to. Swapping renderers is a repo-wide decision, not something to settle in a crash fix.

